PR #66, opened 2021-04-06
Add testing helpers

crynobone:
```php
$responses = \Laravel\Octane\Testing\OctaneRequest::from($this->app)->handle([
    Request::create('/first', 'GET'),
    Request::create('/second', 'GET'),
]);

// $responses[0] and $responses[1] are instances of Illuminate\Testing\TestResponse
```

### Potential issues

* Shouldn't be able to use Event Fake during testing since it would be required to flush states between requests.

Signed-off-by: Mior Muhammad Zaki <crynobone@gmail.com>

### DIFF
--- a/src/Testing/OctaneRequest.php
+++ b/src/Testing/OctaneRequest.php
@@ -8,8 +8,6 @@ use Laravel\Octane\ApplicationFactory;
 use Laravel\Octane\Contracts\Client;
 use Laravel\Octane\Octane;
 use Laravel\Octane\OctaneServiceProvider;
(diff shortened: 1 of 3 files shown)


taylorotwell:
Is the main benefit of this just to try to make sure your application works well with the Octane worker vs. using the typical `$this->get()` type testing helpers?

crynobone:
I'm hoping we can have full access to do similar things to `MakeHttpRequests`. Such as tests uploading file, make JSON requests, use stream response.

taylorotwell:
Going to table this for now.

PR #81, opened 2021-04-06
Remove unused import and fix code style

shuvroroy:


taylorotwell:
Some of these changes are unwanted. We leave those imports in there in case people uncomment them below.

PR #82, opened 2021-04-06
fix code style

shuvroroy:


taylorotwell:
We just let StyleCI handle code style.

PR #87, opened 2021-04-06
Strict compare

nuernbergerA:
Not sure if this was intended


nunomaduro:
Both `==` and `===` are equivalent in this case.

PR #97, opened 2021-04-06
Handle invalid server option in `octane:start/reload/stop`

sebdesign:
For example, if I type `php artisan octane:start --server=swole`, the `roadrunner` server is started.


nunomaduro:
Can you apply the same code to the other commands please?

sebdesign:
Of course!

PR #112, opened 2021-04-07
Use PhpExecutableFinder for Octane server command

barryvdh:
This is similar to the Swoole command, makes sure the current PHP version is used (eg. when using multiple PHP versions)


barryvdh:
I also updated the composer command for Octane, so that composer uses that version used on the commandline for requiring+downloading Octane.

barryvdh:
Haven't checked in other places though.

PR #117, opened 2021-04-07
Fix dispatchProcessSignal, check if the process exists before kill

### LINKED ISSUE #114: Swoole\Process::kill(): swKill(xxxx, 15) failed, Error: No such process[3]
- Octane Version: v0.1.0
- Laravel Version: v8.36.0
- PHP Version: 8.0.3
- Database Driver & Version: Not Related
- Distro: debian:buster (Docker)

** NOT USING LARAVEL SAIL **

### Description:

When using `Ctrl+C` to stop the server the following exception is thrown:
```
 ErrorException

  Swoole\Process::kill(): swKill(7127, 15) failed, Error: No such process[3]

  at vendor/laravel/octane/src/Swoole/SwooleExtension.php:18
     14▕      * @return bool
     15▕      */
     16▕     public function dispatchProcessSignal(int $processId, string $signal): bool
     17▕     {
➜  18▕         return Process::kill($processId, $signal);
     19▕     }
     20▕
     21▕     /**
     22▕      * Set the current process name.

      +45 vendor frames
  46  artisan:37
```

Afterwards the `octane-server-state.json` is still present and it's impossible to start the server again because of:
```
 ERROR  Server is already running.
```

After some research found out that in `Laravel\Octane\Swoole\ServerProcessInspector::stopServer()` if i remove the:
```
$this->dispatcher->terminate($masterProcessId, 15)
```

the server terminates without problems and the `octane-server-state.json` is missing like it's supposed to, and after when checking with `htop` both processes (master & manager) are gone.

### Steps To Reproduce:

I guess trying with a fresh laravel installation within a debian:buster container to run:
```
/usr/bin/php -d variables_order=EGPCS /var/www/app/artisan octane:start --server=swoole --host=0.0.0.0 --port=8080 --workers=1
```

and stop it with `Ctrl+C`.

Happy to help with anything.

sy-records:
fix #114

### DIFF
--- a/src/Swoole/SwooleExtension.php
+++ b/src/Swoole/SwooleExtension.php
@@ -15,7 +15,11 @@ class SwooleExtension
      */
     public function dispatchProcessSignal(int $processId, string $signal): bool
     {
-        return Process::kill($processId, $signal);
+        if (Process::kill($processId, 0)) {
+            return Process::kill($processId, $signal);
+        }
+
+        return false;
     }
 
     /**


src/Swoole/SwooleExtension.php
```diff
@@ -15,7 +15,11 @@ class SwooleExtension
      */
     public function dispatchProcessSignal(int $processId, string $signal): bool
     {
-        return Process::kill($processId, $signal);
+        if (Process::kill($processId, 0)) {
```
Namoshek, line 18: According to [the english docs](https://www.swoole.co.uk/docs/modules/swoole-process-kill), the second parameter expects a string and is optional. Why pass `0` here?

sy-records: There are some problems with the English document. I've contacted to modify it.

Kill 0 is used to detect whether the process exists.

Namoshek: The [PHP extension docs](https://www.php.net/manual/en/swoole-process.kill.php) show the same, although I've no idea how and from what these are generated. Just pointing it out though, not saying yours is wrong.

PR #125, opened 2021-04-07
Multiple os testing setup

rogervila:
Run GitHub actions on Linux and OSX. 

Windows is disabled due to its lack of signal handling support.


taylorotwell:
We don't test our other packages on Mac. /cc @driesvints

driesvints:
We only test on Ubuntu for most of our first party packages. But I'll put it on my list to eventually also check into Windows builds for some of them. I can see it being useful for Octane.

rogervila:
Thank you @driesvints 
I can open the same PR for other Laravel projects if you find it useful.

PR #132, opened 2021-04-07
Drop realpath to serve linked files

### LINKED ISSUE #99: Unable to serve static resources from a symink folder.
- Octane Version: beta
- Laravel Version: 8.35.1
- PHP Version: 8.0.3

### Description:
Octane is unable to serve static resources configured through a symlink.  I have a separate node project where I keep my static resource (js & css files) I then use a symlink to reference these from the laravel's public folder
`/home/naveed/laravel/public/assets` -> `/home/naveed/ui/build/assets/`  

Octane is using `realpath` which resolves the symlink thus making it incomptible to render static resources. `http://127.0.0.1:8000/assets/main.css`  that results in a 404 Not Found.

```php
  public function canServeRequestAsStaticFile(Request $request, RequestContext $context): bool
    {
        if (! ($context->publicPath ?? false) ||
            $request->path() === '/') {
            return false;
        }

        $publicPath = $context->publicPath;

        return $this->fileIsServable(
            $publicPath,
            realpath($publicPath.'/'.$request->path()),
        );
    }
```
Removing the `realpath` solves the problem but I am not sure if it has any security implications.

ahmedash95:
This PR drops usage of `realpath() in favor of serving linked files. fixes #99 

I'm not sure what to do with the testcase. as this change would allow to read files out of public directory. and `is_link` only works if the file itself is symlinked. but in #99 case it's the parent directory that is symlinked. 

so up to the octane team to decide on that

### DIFF
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -53,7 +53,7 @@ class SwooleClient implements Client, ServesStaticFiles
 
         return $this->fileIsServable(
             $publicPath,
-            realpath($publicPath.'/'.$request->path()),
+            $publicPath.'/'.$request->path()
         );
     }
 
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -53,7 +53,7 @@ class SwooleClient implements Client, ServesStaticFiles
 
         return $this->fileIsServable(
             $publicPath,
-            $publicPath.'/'.$request->path()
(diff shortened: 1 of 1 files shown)


driesvints:
@ahmedash95 tests are failing

ahmedash95:
> @ahmedash95 tests are failing

Since it's not easy to drop realpath. I'm gonna work on better alternative this night

driesvints:
@ahmedash95 feel free to re-attempt the PR 👍

PR #134, opened 2021-04-08
Ignore .idea

sy-records:
Ignore `.idea` when developing with PhpStorm


nunomaduro:
We already have those in our system `.gitignore` file.

PR #135, opened 2021-04-08
Add swoole configuration options

### LINKED ISSUE #98: Missing config option for swoole server set?
When looking at 
https://github.com/laravel/octane/blob/master/bin/createSwooleServer.php#L20

Then are we not missing the following in the octane config? 

```php
'swoole' => [
    'options' => [
         // see https://www.swoole.co.uk/docs/modules/swoole-server/configuration
    ],
],
```

It would be great to be able to pass config to the swoole server set feature from the config/octane.php.

sy-records:
While using Swoole, you can set additional configuration options as needed.

close #98

### DIFF
--- a/config/octane.php
+++ b/config/octane.php
@@ -180,4 +180,20 @@ return [
         ],
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Swoole configuration options
+    |--------------------------------------------------------------------------
+    |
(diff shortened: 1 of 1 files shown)


taylorotwell:
I may document this later as the beta moves on but for right now I'm keeping the configuration file simple.

PR #138, opened 2021-04-08
Optimized the memory usage when writing chunks

hhxsv5:
Hello guys, I did a test, substr() has no memory overhead and is more efficient than str_split().

```php
$bigStr = str_repeat('abcdefghijklmnopqrstuvwxyz0123456789', 1000000);
$memStart = memory_get_usage(true);
$timeStart = microtime(true);

// ----- substr()
$len = strlen($bigStr);
$chunkLimit = 4096;
for ($offset = 0; $offset < $len; $offset += $chunkLimit) {
    $chunk = substr($bigStr, $offset, $chunkLimit);
}

// ----- str_split()
//$chunkLimit = 4096;
//foreach (str_split($bigStr, $chunkLimit) as $chunk) {
//}

$costMem = memory_get_usage(true) - $memStart;
$costTime = microtime(true) - $timeStart;
echo "Memory: {$costMem} Time: {$costTime}\n";
```

```
// substr()
Memory: 0 Time: 0.0036599636077881

// str_split()
Memory: 2097152 Time: 0.043421983718872
```
[reference](https://github.com/hhxsv5/laravel-s/blob/master/src/Swoole/DynamicResponse.php#L31)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 composer.lock
 /phpunit.xml
 .phpunit.result.cache
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,3 @@
 composer.lock
 /phpunit.xml
 .phpunit.result.cache
-.idea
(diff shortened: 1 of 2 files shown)


barryvdh:
Why is your reference using factor 0.6 upon writing? Because of gzipping?
And why is the chunkSize 2MB instead of one there?

PR #144, opened 2021-04-08
Serve symlinks

ahmedash95:
This PR solves the #99 of serving symlinked files

### DIFF
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -50,11 +50,39 @@ class SwooleClient implements Client, ServesStaticFiles
         }
 
         $publicPath = $context->publicPath;
+        $fullPath = $publicPath.'/'.$request->path();
+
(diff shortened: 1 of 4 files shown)


taylorotwell:
Does this protect again file system traversal or being able to access any file on the filesystem?

`http://localhost/something/../../../.env` or something similar for example?

ahmedash95:
Since we check for symlinks only, accessing other parent directories like ../.env will not work. and there is already a test for that https://github.com/laravel/octane/blob/master/tests/SwooleClientTest.php#L71.

So the drop to realpath only happens if there is a symlink.

Hope that answered your questions.

taylorotwell:
This PR allows the retrieval of *any* file on the filesystem accessible by the user running Octane. Run this from your application:

```bash
php artisan storage:link

curl --path-as-is 'http://127.0.0.1:8000/storage/../../../.env'
```

Issue #109, opened 2021-04-07
Response doesn't support stream-based output


albertcht:
- Octane Version: 0.1.0
- Laravel Version: 8.36.1
- PHP Version: 8.0.3
- Database Driver & Version:

### Description:

Most of PHP's output usages like `echo`, `var_dump` output its content to buffer. In Swoole's case the result will print on console instead of web.

It seems this partial code in `SwooleClient.php` refers from [here](https://github.com/swooletw/laravel-swoole/blob/3cf0cd997d0503f1cc73c8d330818b81dadd6f95/src/Transformers/Response.php#L110-L122) https://github.com/laravel/octane/blob/08537a7ed6d2c463f76278081ac7af9a6dd1dead/src/Swoole/SwooleClient.php#L153-L178 because the `output` property doesn't exist in symfony's response object. The `output` property is set from [here](https://github.com/swooletw/laravel-swoole/blob/3cf0cd997d0503f1cc73c8d330818b81dadd6f95/src/Server/Sandbox.php#L236) in `swooletw/laravel-swoole`.

`Symfony\Component\HttpFoundation\StreamedResponse` doesn't support `getContent` and `setContent` for setting output content, so I fetch buffer output  and set the result to response's `output` property manually.

If Octane is going to support stream output, it needs to use `ob_` functions to catch stream content as well.

Many thanks to the official team for integrating Swoole into Laravel!

### Steps To Reproduce:

```php
echo 'hello world';

var_dump('Swoole is awesome');

response()->streamDownload(function () {
    echo file_get_contents("https://github.com/laravel/laravel/blob/8.x/README.md");
    }, 'laravel-readme.md');
});
```

themsaid:
@albertcht can you open a PR to show the changes needed?

albertcht:
> @albertcht can you open a PR to show the changes needed?

Sure, I will submit a PR to show the changes needed for stream-based output.

27pchrisl:
It would be ideal if there were a way to access the current `Swoole\Http\Response`, so that I could call `write()` on it.

My use case is streaming large JSON objects, which currently is done using a JSON generator calling `echo` within the `StreamResponse::sendContent` callback. These objects are too large to hold in memory, so a way of doing this that doesn't involve output buffering would be great!

27pchrisl:
[...]

PR #151, opened 2021-04-09
[Swoole] Send StreamedResponse

barryvdh:
Capture the output from a StreamdResponse.  Fixes https://github.com/laravel/octane/issues/109

Note that this doesn't actually stream the contents anymore, but it does at least render the output (so after the callback is finished).

### DIFF
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -156,11 +156,7 @@ class SwooleClient implements Client, ServesStaticFiles
      */
     protected function sendResponseContent(Response $response, SwooleResponse $swooleResponse): void
     {
-        if ($response instanceof StreamedResponse && property_exists($response, 'output')) {
-            $swooleResponse->end($response->output);
-
-            return;
-        } elseif ($response instanceof BinaryFileResponse) {
+        if ($response instanceof BinaryFileResponse) {
             $swooleResponse->sendfile($response->getFile()->getPathname());
 
             return;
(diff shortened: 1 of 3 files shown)


themsaid:
@barryvdh what do you think of https://github.com/laravel/octane/pull/153 ?

barryvdh:
So capture all output? Technically I think that a proper response object is always the way to go, but I understand the appeal of dump(), and it's closer to 'normal'

barryvdh:
So based on the comment, the previous if statement doesn't actually do anything ever. I couldn't really see where the `output` property was coming from, but as I gather from https://github.com/laravel/octane/issues/109#issue-852101428 that based on laravel-swoole, who do use that. I think it's safe to remove here?

barryvdh:
(RoadRunner was already working, the PSR-7 factory does something similar, so just added a test)

barryvdh:
(RoadRunner was already working, the PSR-7 factory does something similar, so just added a test)

albertcht:
> So based on the comment, the previous if statement doesn't actually do anything ever. I couldn't really see where the `output` property was coming from, but as I gather from [#109 (comment)](https://github.com/laravel/octane/issues/109#issue-852101428) that based on laravel-swoole, who do use that. I think it's safe to remove here?

I think removing the first statement for stream response is safe.

taylorotwell:
Will follow up with this on #153 ... thanks!

Radiergummi:
Hey there @barryvdh , 

> Note that this doesn't actually stream the contents anymore, but it does at least render the output (so after the callback is finished).

So, I tried to follow the trails through the issues but couldn't find directions or documentation on stream responses using Octane. Is this generally impossible? If so, that would make a great addition to the documentation, and it's a real headache for our application :( Is there any recommended strategy, any idea?  
We have this data export endpoint that streams hundreds of megabytes of CSV reports...

PR #153, opened 2021-04-09
Collect output buffer so echo, var_dump(), and dump() work

themsaid:
This PR collects the output buffer and prepends it to the response.

It also handles StreamedResponses as reported in https://github.com/laravel/octane/issues/109. However, It'll collect the response and send it at the end of the request, it won't **stream** it.

Screen Shot 2021-04-09 at 12 52 02 PM

### DIFF
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -169,7 +169,7 @@ class SwooleClient implements Client, ServesStaticFiles
 
         if ($octaneResponse->response instanceof StreamedResponse) {
             ob_start(function ($data) use ($swooleResponse) {
-                logger('write ' . strlen($data));
+                logger('write '.strlen($data));
                 if (strlen($data) > 0) {
                     $swooleResponse->write($data);
                 }
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -169,12 +169,13 @@ class SwooleClient implements Client, ServesStaticFiles
 
         if ($octaneResponse->response instanceof StreamedResponse) {
             ob_start(function ($data) use ($swooleResponse) {
+                logger('write ' . strlen($data));
                 if (strlen($data) > 0) {
                     $swooleResponse->write($data);
                 }
 
                 return '';
-            }, 10);
+            }, 1);
 
             $octaneResponse->response->sendContent();
 
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -169,7 +169,7 @@ class SwooleClient implements Client, ServesStaticFiles
 
         if ($octaneResponse->response instanceof StreamedResponse) {
             ob_start(function ($data) use ($swooleResponse) {
-                if ($data) {
+                if (strlen($data) > 0) {
                     $swooleResponse->write($data);
                 }
 
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -167,7 +167,17 @@ class SwooleClient implements Client, ServesStaticFiles
             return;
         }
 
-        $content = $octaneResponse->response->getContent();
+        if ($octaneResponse->response instanceof StreamedResponse) {
+            ob_start();
+
+            $octaneResponse->response->sendContent();
+
+            $content = ob_get_contents();
+
+            ob_end_clean();
+        } else {
+            $content = $octaneResponse->response->getContent();
+        }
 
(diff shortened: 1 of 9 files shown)


albertcht:
Thanks for the quick action!

barryvdh:
This does catch more cases I suppose. Although technically it would be prettier to use proper response objects, but it does make it easier. Maybe copy the tests from https://github.com/laravel/octane/pull/153 to verify the streams are working (without streaming, but at least working)?

src/Swoole/SwooleClient.php
```diff
@@ -150,23 +151,37 @@ public function sendResponseHeaders(Response $response, SwooleResponse $swooleRe
...
+     * @param  \Laravel\Octane\OctaneResponse  $response
      * @param  \Swoole\Http\Response  $response
      * @return void
      */
-    protected function sendResponseContent(Response $response, SwooleResponse $swooleResponse): void
+    protected function sendResponseContent(OctaneResponse $octaneResponse, SwooleResponse $swooleResponse): void
     {
-        if ($response instanceof StreamedResponse && property_exists($response, 'output')) {
-            $swooleResponse->end($response->output);
+        if ($octaneResponse->response instanceof StreamedResponse && property_exists($octaneResponse->response, 'output')) {
```
barryvdh, line 160: I think this case is obsolete and never used? The output property is never set in Octane, as discussed in https://github.com/laravel/octane/issues/109

barryvdh:
Ah I was trying the callback function for `ob_start`, but missed the chunk option 🙄

src/Swoole/SwooleClient.php
```diff
@@ -157,32 +157,36 @@ public function sendResponseHeaders(Response $response, SwooleResponse $swooleRe
...
         }
 
+        if ($octaneResponse->outputBuffer) {
+            $swooleResponse->write($octaneResponse->outputBuffer);
+        }
+
         if ($octaneResponse->response instanceof StreamedResponse) {
-            ob_start();
+            ob_start(function ($data) use ($swooleResponse) {
+                if ($data) {
```
barryvdh, line 172: I would suggest `strlen($data) > 0` because it can be " " or "0" etc

src/Swoole/SwooleClient.php
```diff
@@ -157,32 +157,36 @@ public function sendResponseHeaders(Response $response, SwooleResponse $swooleRe
...
+                    $swooleResponse->write($data);
+                }
 
-            $octaneResponse->response->sendContent();
+                return '';
+            }, 10);
```
barryvdh, line 177: Is `1` safer to use? Bit edge case but if you have something like:
```php
echo '1';
sleep(5);
echo '2';
```
It doesn't return directly.

crynobone: https://user-images.githubusercontent.com/172966/114196759-3f336e00-9984-11eb-9410-0d7ac41733c7.mp4

Here's some test with `10`, not sure if `1` (byte) suitable for chunk. Maybe let's everyone try first?

barryvdh: I tested with chunksize 1:

```php
    return \response()->stream(function(){

       echo str_repeat('T', random_int(1, 1000));
       sleep(2);
        echo str_repeat('E', random_int(1, 1000));
        sleep(2);
        echo str_repeat('S', random_int(1, 1000));
        sleep(2);
        echo str_repeat('T', random_int(1, 1000));
        sleep(2);
       echo 'done';
    });
```
And added some logging to the ob callback:
```php
ob_start(function($data) use($swooleResponse) {
                logger('write ' . strlen($data));
                if (strlen($data)) {
                    $swooleResponse->write($data);
                }

                return '';
            }, 1);
```

It still chunks to parts for me:
```
[2021-04-09 14:45:35] local.DEBUG: write 676  
[2021-04-09 14:45:37] local.DEBUG: write 601  
[2021-04-09 14:45:39] local.DEBUG: write 492  
[2021-04-09 14:45:41] local.DEBUG: write 4  
[2021-04-09 14:45:41] local.DEBUG: write 0  
```

So with chunkSize 10 I think it's possible that small parts are stuck untill a new item is flushed.

taylorotwell:
@themsaid was any more follow up needed on @barryvdh's comments?

themsaid:
@barryvdh @crynobone @taylorotwell I updated the chunk size to 1

barryvdh:
Nice! I've added the tests from my previous PR which also checks that write() is called once; https://github.com/laravel/octane/pull/157

PR #159, opened 2021-04-10
Add swoole/ide-helper to require-dev

Yurunsoft:
This package contains IDE help files for Swoole. You may use it in your IDE to provide accurate autocompletion.

This package is officially released by Swoole.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,8 @@
         "orchestra/testbench": "^6.16",
         "phpunit/phpunit": "^9.3",
         "spatie/laravel-ray": "^1.14",
-        "spiral/roadrunner": "^2.0"
+        "spiral/roadrunner": "^2.0",
+        "swoole/ide-helper": "^4.6"
     },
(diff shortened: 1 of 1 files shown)


crynobone:
Shouldn't this be as suggested package instead?

barryvdh:
I think it isn't as useful for end-users, but more for developers right? I noticed that PhpStorm didn't understand the Swoole functions, so this could be helpful while developing Octane.

Yurunsoft:
> I think it isn't as useful for end-users, but more for developers right? I noticed that PhpStorm didn't understand the Swoole functions, so this could be helpful while developing Octane.

Yes, you're right

themsaid:
I tried it and it didn't work for me on PHPStorm.

Yurunsoft:
> I tried it and it didn't work for me on PHPStorm.

I use vscode and it works well.

swoole/ide-helper also support phpstorm.

image

PR #165, opened 2021-04-10
Typo

rochdiNassah:

### DIFF
--- a/src/MarshalsPsr7RequestsAndResponses.php
+++ b/src/MarshalsPsr7RequestsAndResponses.php
@@ -58,7 +58,7 @@ trait MarshalsPsr7RequestsAndResponses
      *
      * This instance can turn a PSR-7 request into an HttpFoundation request.
      *
-     * @return \Symfony\Bridge\PsrHttpMessage\Factory\HttpFoundationFactoryInterface
+     * @return \Symfony\Bridge\PsrHttpMessage\HttpFoundationFactoryInterface
      */
     protected function httpFoundationRequestFactory(): HttpFoundationFactoryInterface
     {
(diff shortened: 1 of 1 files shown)


src/MarshalsPsr7RequestsAndResponses.php
```diff
@@ -58,7 +58,7 @@ protected function toPsr7Response(Response $response): ResponseInterface
      *
      * This instance can turn a PSR-7 request into an HttpFoundation request.
      *
-     * @return \Symfony\Bridge\PsrHttpMessage\Factory\HttpFoundationFactory
+     * @return \Symfony\Bridge\PsrHttpMessage\Factory\HttpFoundationFactoryInterface
```
nunomaduro, line 61: Maybe is `Symfony\Bridge\PsrHttpMessage\HttpFoundationFactoryInterface`?

nunomaduro (changes requested):

src/MarshalsPsr7RequestsAndResponses.php
rochdiNassah: It is already changed to ```HttpFoundationFactoryInterface```.

nunomaduro: But the FQN is not correct. Check this: https://github.com/symfony/psr-http-message-bridge/blob/main/HttpFoundationFactoryInterface.php.

rochdiNassah: I checked and it seems correct as the last change.

Namoshek: No, it is not correct. The `HttpFoundationFactoryInterface` is imported as `Symfony\Bridge\PsrHttpMessage\HttpFoundationFactoryInterface` in the same file on line 14:

https://github.com/laravel/octane/blob/dc393ebf101c880d4b96c5560d2d50bea6adb91a/src/MarshalsPsr7RequestsAndResponses.php#L14

rochdiNassah: Changed, yes you're right.
thanks!

PR #168, opened 2021-04-10
Allow for swoole to serve static files through symlink

maxeckel:
This PR resolves issue #99 and is based on the PR by @ahmedash95 #144.

The PR resolves the issue of being able to use directory traversal in conjunction with symlinks mentioned by @taylorotwell in #144.

In order to prevent the directory traversal, I take the path following the symlink and check whether the resolved path using realpath() ends with the requested path.

**For example:**

Requested URL: `http://127.0.0.1:8000/storage/../../../.env`

realpath() output: `/my/project/directory/.env`

Since `storage` is the symlink, the path `../../../.env` will be checked against the realpath() output. 
Because of the traversal attempt, the paths don't match and the request won't be served statically.


src/Swoole/SwooleClient.php
```diff
@@ -56,17 +56,67 @@ public function canServeRequestAsStaticFile(Request $request, RequestContext $co
...
+
+    /**
+     * Determine whether the path contains a symlink.
+     * When a symlink is found, the path after it is returned.
+     *
+     * @param  string $publicPath
+     * @param  string $path
+     * @return string|bool
+     */
+    private function pathContainsSymlink(string $publicPath, string $path): string | bool
```
GrahamCampbell, line 100: Don't use union types in returns. StyleCI doesn't format them correctly (yet).

GrahamCampbell (changes requested):

src/Swoole/SwooleClient.php
maxeckel: Thanks Graham, it has been removed.

barryvdh:
@maxeckel Can you check https://github.com/laravel/octane/pull/179 
I think the test was wrong but unintentionally passed for a while.

PR #178, opened 2021-04-12
Reload must run, otherwise throw error

barryvdh:
If the process exits, throw an error so it's more clear.
In my case it says this, but not sure if that's just me. At least for now show that it fails.

```
  INFO  Reloading workers...

   Symfony\Component\Process\Exception\ProcessFailedException 

  The command "'./rr' 'reset'" failed.

Exit Code: 1(General error)

Working directory: /home/barry/www/octane

Output:
================


Error Output:
================
rpc service disabled

```

### DIFF
--- a/src/RoadRunner/ServerProcessInspector.php
+++ b/src/RoadRunner/ServerProcessInspector.php
@@ -39,7 +39,7 @@ class ServerProcessInspector
     {
         $this->processFactory->createProcess([
             './rr', 'reset',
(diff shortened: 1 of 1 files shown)


barryvdh:
Maybe the reloadServer should return true/false on success/error, or always throw an exception when fails (so Swoole also)?
If this is part of a deployment script, you would perhaps want to abort when reload fails.

PR #179, opened 2021-04-12
Make sure file is not empty, test outside public

barryvdh:
I think the test in https://github.com/laravel/octane/pull/168 is probably not correct, as the result of the files leads to a file inside the public folder, so it SHOULD be accessible?
The reason the test was working was that the reference file was empty. But since https://github.com/laravel/octane/pull/162 was merged, this caused the test to fail suddenly. 

This PR:
 - Makes sure the bar.txt is not empty. Although it's not used anymore so might be removed?
 - Links to a file outside the public dir, to test the code.

Not totally sure if this is 100% secure now though, but I think this was the intent of the previous PR?


maxeckel:
Hey @barryvdh,

I've checked your concerns and you are right.

The symlink is resolved correctly and forbids access through the directory traversal. But since the file is within the public directory, the call to `realpath()` actually resolves the path correctly and therefore the file is delivered and the test fails.

So your fix to check against the file outside of the public folder is totally correct. Thank you! I'm sorry I missed this one...

PR #180, opened 2021-04-12
Refactor writeProcessId and writeState.

eleven26:
It seems too many duplicate codes, and there may be other states to write to state file in the future.


taylorotwell:
We will address any code duplications, etc. when the situation arises.

Issue #171, opened 2021-04-11
memory leak


caiking:
- Octane Version:0.1.0
- Laravel Version: 8.36.2
- PHP Version: 8.0.3
- Server: Swoole-4.6.5

### Description:

i just install laravel,swoole,octane

and start with "php artisan octane:start --port=13000 --watch"

add code like this:

```
<?php


namespace App\Http\Controllers;


class TestController extends Controller
{
    public function index()
    {
        return memory_get_usage()/1024/1204;
    }
}
```

and this:

`Route::get('/test', App\Http\Controllers\TestController::class."@index");`

then did nothing

and i requested "http://127.0.0.1:13000/api/test" in chrome ，often and often

i got these respopnse:
13.533650851329
13.710976432724
13.747858700166 
 ...
 15.084081706811
 ...

The numbers are not going down

Namoshek:
Try configuring a lower threshold for garbage collection in `config/octane.php`.

albertcht:
Sometimes it's difficult to track memory leak in your Laravel project in long-running environment. In traditional PHP lifecycle, developers don't need to worry about it because all the resources will be destroyed after each request.

There might be somewhere using global variable wrongly in your project (maybe in other packages). You can try to add `max-requests` argument when you start the server. And workers will be gracefully restarted after serving that amount of requests to avoid memory leaks.

```
php artisan octane:start --max-requests=250
```

themsaid:
Investigating this. Thanks for reporting.

linaspasv:
I have experienced the same on a fresh laravel install.

themsaid:
fixed here https://github.com/laravel/octane/pull/181

kocoten1992:
Hi @themsaid, how you was able detect memory leak?

I got a leak and was trying to find a memory leak detector for swoole, but find none so far.

PR #181, opened 2021-04-12
Fix memory leaks

themsaid:
This PR fixes memory leaks in Octane as reported in https://github.com/laravel/octane/issues/171

---

```
$sandbox->flush();
```

This will flush the entire sandbox container. This will clear any references to request-content singletons which allows PHP's garbage collector from clearing the memory when `unset($sandbox)` is called.

---

```
$this->app->make('view.engine.resolver')->forget('blade');
$this->app->make('view.engine.resolver')->forget('php');
```

These lines deal with a memory leak in facade/ignition as explained in https://github.com/laravel/framework/pull/36955.

### DIFF
--- a/src/Worker.php
+++ b/src/Worker.php
@@ -4,10 +4,8 @@ namespace Laravel\Octane;
 
 use Closure;
 use Illuminate\Container\Container;
-use Illuminate\Contracts\Auth\Access\Gate;
 use Illuminate\Foundation\Application;
 use Illuminate\Http\Request;
-use Illuminate\Session\Middleware\StartSession;
 use Laravel\Octane\Contracts\Client;
 use Laravel\Octane\Contracts\ServesStaticFiles;
 use Laravel\Octane\Contracts\Worker as WorkerContract;
--- a/src/Worker.php
+++ b/src/Worker.php
@@ -3,8 +3,11 @@
 namespace Laravel\Octane;
 
 use Closure;
+use Illuminate\Container\Container;
+use Illuminate\Contracts\Auth\Access\Gate;
 use Illuminate\Foundation\Application;
 use Illuminate\Http\Request;
+use Illuminate\Session\Middleware\StartSession;
 use Laravel\Octane\Contracts\Client;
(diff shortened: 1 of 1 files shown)


taylorotwell:
@themsaid looks like our tests need fixing.

themsaid:
@taylorotwell the PR depends on https://github.com/laravel/framework/pull/36955 so it'll be fixed when we can lock to the release tomorrow.

src/Worker.php
```diff
@@ -110,6 +113,15 @@ public function handle(Request $request, RequestContext $context): void
         } catch (Throwable $e) {
             $this->handleWorkerError($e, $sandbox, $request, $context, $responded);
         } finally {
+            $sandbox->forgetInstance('request');
```
albertcht, line 116: How about collecting these instances in `FlushTemporaryContainerInstances` listener?

themsaid: I've changed the code to flush everything.

src/Worker.php
```diff
@@ -110,6 +111,11 @@ public function handle(Request $request, RequestContext $context): void
         } catch (Throwable $e) {
             $this->handleWorkerError($e, $sandbox, $request, $context, $responded);
         } finally {
+            $sandbox->flush();
+
+            $this->app->make('view.engine.resolver')->forget('blade');
```
GrahamCampbell, line 116: Is there a way to forget all? If not, could it be added to Laravel today. Otherwise, other custom things will still cause leaks.

GrahamCampbell (changes requested):

src/Worker.php
```diff
@@ -110,6 +111,11 @@ public function handle(Request $request, RequestContext $context): void
         } catch (Throwable $e) {
             $this->handleWorkerError($e, $sandbox, $request, $context, $responded);
         } finally {
+            $sandbox->flush();
```
albertcht, line 114: Then it's okay to remove other sandbox' forgetInstance call in FlushTemporaryContainerInstances?
(Not sure if it's necessary to call forgetInstance in FlushAuthenticationState)

PR #184, opened 2021-04-12
Allow reuse the octane cache after stop the swoole server

mohamed-samir907:
This PR allow reuse the octane cache after the swoole server stop.
---
The issue is when we stop the swoole server the cached data will be flushed.

To solve that i stored the cached data when we call the command `php artisan octane:stop` into the default cache driver.

And on `managerstart` event we restore this cached data and put it again in swoole cache table.

```php
Route::get('/', function () {

    var_dump(Cache::store('octane')->get('name'));

    Cache::store('octane')->put('name', 'mohamed samir');

    return view('welcome');
});

```
The first time `var_dump(Cache::store('octane')->get('name'));` will return `null` because we don't have any cached value. Other than that we can get the cached value, even if we stopped the server.

You can try this example
1. run `php artisan octane:start`
2. open `http://localhost:8000` you will see `null`. refresh the page you will see the cached value.
3. run `php artisan octane:stop` then `php artisan octane:start`
4. open `http://localhost:8000` you will see the cached value.

### DIFF
--- a/bin/swoole-server
+++ b/bin/swoole-server
@@ -3,6 +3,7 @@
 
 use Laravel\Octane\RequestContext;
 use Laravel\Octane\Swoole\Handlers\OnManagerStart;
+use Laravel\Octane\Swoole\Handlers\OnManagerStop;
 use Laravel\Octane\Swoole\Handlers\OnServerShutdown;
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #185, opened 2021-04-12
Add state_file in config

rennokki:


barryvdh:
Why?

rennokki:
@barryvdh To be able to change the state file location on some environments: https://github.com/laravel/octane/blob/master/src/OctaneServiceProvider.php#L53-L56

PR #195, opened 2021-04-14
Handle empty responses

themsaid:
Without this, the following exception is thrown.

Screen Shot 2021-04-14 at 11 54 47 AM

### DIFF
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -230,7 +230,7 @@ class SwooleClient implements Client, ServesStaticFiles
 
         $content = $octaneResponse->response->getContent();
 
-        if($length = strlen($content) === 0){
+        if ($length = strlen($content) === 0) {
             $swooleResponse->end();
 
             return;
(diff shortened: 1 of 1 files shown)


src/Swoole/SwooleClient.php
```diff
@@ -230,7 +230,11 @@ protected function sendResponseContent(OctaneResponse $octaneResponse, SwooleRes
 
         $content = $octaneResponse->response->getContent();
 
-        $length = strlen($content);
+        if ($length = strlen($content) === 0) {
```
cerbero90, line 233:
```suggestion
        if (($length = strlen($content)) === 0) {
```

without wrapping parenthesis, `$length` would become a boolean

nunomaduro:
@themsaid Can you add a test on this? Let me know if you want me to add a test for you.

src/Swoole/SwooleClient.php
barryvdh: I would just leave it a 2-liner :P

Issue #190, opened 2021-04-13
Octane::concurrently may cause requests to be unresponsive


eleven26:
- Octane Version: v0.1.1
- Laravel Version:  v8.36.2
- PHP Version:  8.0.3
- Server: Swoole 
- Swoole Version: 4.6.5
- Database Driver & Version: MySQL 5.7.26

### Description:

When multiple MySQL queries are launched in `Octane:: concurrently`, the worker process may not respond.

#### Two tasks can respond normally

I'd like to try to run multiple MySQL queries at the same time, and then use `Octane:: concurrently` to test. The test code is as follows:

```PHP
Route::get('/', function () {
    $result = Octane::concurrently([
        function() {
            return DB::select(DB::raw('select sleep(1);'));
        },
        function() {
            return DB::select(DB::raw('select sleep(1);'));
        },
    ]);

    var_dump($result);

    return view('welcome');
});
```

When requested, two MySQL query tasks are initiated. When I made the request, the server returned normally.

#### More MySQL queries are unresponsive

Then I increased the number of tasks by one, modified the query statement, and monitored the `TaskTerminated` event：

```PHP
Event::listen(TaskTerminated::class, function () {
    echo "Task terminated";
});

Route::get('/', function () {
    $result = Octane::concurrently([
        function() {
            return DB::select(DB::raw('select sleep(1);'));
        },
        function() {
            return DB::select(DB::raw('select sleep(2);'));
        },
        function() {
            return DB::select(DB::raw('select sleep(3);'));
        },
]);

    var_dump($result);

    return view('welcome');
});

```

But when I started the request, the server didn't respond at all. It was always in the pending state. But you can see that the console outputs three `Task terminated`.

I used laravel sail to test. I went in and used `ps -ef` to look at the process status. I found that there are two manager processes, but normally there should be only one.

```
sail        26    19  0 16:14 ?        00:00:00 swoole_http_server: master process for Laravel
sail        27    26  0 16:14 ?        00:00:00 swoole_http_server: manager process for Laravel
sail        30    27  0 16:14 ?        00:00:00 /usr/bin/php8.0 swoole-server /var/www/html/storage/logs/octane-server-state.json
sail        31    27  0 16:14 ?        00:00:00 /usr/bin/php8.0 swoole-server /var/www/html/storage/logs/octane-server-state.json
sail        32    27  0 16:14 ?        00:00:00 /usr/bin/php8.0 swoole-server /var/www/html/storage/logs/octane-server-state.json
[...]

PR #200, opened 2021-04-15
Document concurrently gotcha

driesvints:
I believe it's important that we document these gotchas so people know about them and we prevent duplicate issues from popping up.

See https://github.com/laravel/octane/issues/190

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,6 +409,12 @@ return Octane::table('example')->get('uuid');
 
 **Note:** The column types supported by Swoole tables are: `string`, `int`, and `float`.
 
+### FAQ
+
+#### `Octane::concurrently` may cause requests to be unresponsive
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
What is the issue that you are trying to describe? Are you referring to if the concurrent tasks take a very long time to execute?

driesvints:
@taylorotwell I just copied @themsaid's reply on the issue.

taylorotwell:
OK - I guess we'll need to wait for follow up from @themsaid on that for more detail.

themsaid:
It's not related to `Octane::concurrently` specifically. Any request that reaches the timeout will hand after its worker is killed.

taylorotwell:
Don't feel like we have quite firmed this up yet in terms of what we are actually warning people against.

PR #208, opened 2021-04-16
Fixed Swoole\Http\Response::write(): data to send is empty

lilianjin:


Namoshek:
How can this even occur with the if above?

barryvdh:
This was fixed with https://github.com/laravel/octane/pull/195 but not yet patched. It occurs when you don't send any response.

driesvints:
Thanks @barryvdh

driesvints:
Patches are on Tuesday

PR #209, opened 2021-04-16
Fixes starting multiple roadrunner servers

nunomaduro:
This pull request reverts:  https://github.com/laravel/octane/pull/149.

If multiple roadrunner servers are running in the same server, the `octane:start` command will fail because of:

```
   ERROR  2021-04-16T12:11:38.712+0100	ERROR	container/poller.go:16	vertex got an error	{"vertex id": "status.Plugin", "error": "listen tcp4 127.0.0.1:2114: bind: address already in use"}
```

### DIFF
--- a/src/Commands/stubs/rr.yaml
+++ b/src/Commands/stubs/rr.yaml
@@ -1,4 +1,0 @@
-rpc:
-  listen: tcp://127.0.0.1:6001
-status:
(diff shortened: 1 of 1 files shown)


litan1106:
Good. we can still use custom .rr.yaml in the project root for rr cli commands.

image

PR #219, opened 2021-04-20
Removes "We are still in the process of ensuring Octane compatibility with Laravel packages"

nunomaduro:
I guess this is no longer the case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Laravel Octane supercharges your application's performance by serving your appli
 
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


driesvints (approved):

Namoshek:
Would it be reasonable to replace this with a hint where potential issues of first- and third-party packages with Octane should be reported? From what I understand, first-party package support has been ensured by now, but since Octane is still in a beta, issue reporting is very important on the one hand and maybe a bit confusing on the other.

driesvints:
@Namoshek on Octane repo itself. But yeah a note could help. @nunomaduro maybe add that one?

PR #224, opened 2021-04-21
Use constants instead of hard coding

sy-records:
`SWOOLE_LOG_TRACE` => 1
`SWOOLE_LOG_INFO` => 2
`SWOOLE_LOG_ERROR` => 5

When using `1` you need to add the `-enable-trace-log` option to compile Swoole, Default is `2`.

### DIFF
--- a/src/Commands/StartSwooleCommand.php
+++ b/src/Commands/StartSwooleCommand.php
@@ -119,7 +119,7 @@ class StartSwooleCommand extends Command implements SignalableCommandInterface
             'enable_coroutine' => false,
             'daemonize' => false,
             'log_file' => storage_path('logs/swoole_http.log'),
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
👎 This hard crashes when swool is not loaded.

huanghantao:
If you haven't installed Swoole, why do you need to run this line of code?

sy-records:
Here there is a determination of whether the extension is installed or not

https://github.com/laravel/octane/blob/a599162ea7e6da8c4e67088e8f2950d8596ad341/src/Commands/StartSwooleCommand.php#L57-L61

GrahamCampbell:
Yes, but that code runs AFTER the code that reads the constants.

sy-records:
image

GrahamCampbell:
Oh, my bad then!

PR #233, opened 2021-04-22
Use extension_loaded instead of function_exists

sy-records:
If the Swoole extension is not loaded, but there is code in other composer packages such as:

```php
var_dump(extension_loaded('swoole'));

if (!function_exists('swoole_cpu_num')) {
    function swoole_cpu_num()
    {
        return 1;
    }
}

var_dump(function_exists('swoole_cpu_num'));
//bool(false)
//bool(true)
```

This creates an error.


GrahamCampbell:
Why don't we want to allow polyfilled swoole?

PR #239, opened 2021-04-23
Use signal method instead of dispatchProcessSignal method

sy-records:
It is not necessary to call `canCommunicateWith`, Judgment exists in the `dispatchProcessSignal` method, so that the modified behavior is consistent.

https://github.com/laravel/octane/blob/e4a02d62b649f4eec24d677bd6101e8f820b49af/src/Swoole/SwooleExtension.php#L26-L33


taylorotwell:
If nothing is broken I will leave it alone.

PR #279, opened 2021-05-17
Fixed rpc-port default port in RoadRunner start command

goszowski:
The default port strictly depends on running the application on port 8000, which is incorrect.


taylorotwell:
Just pass the RPC port you want if you don't like the default behavior.

goszowski:
Default installation with Laravel Sail hast 80 port (Copy Paste from docs). Default installation has PRC port 80 - 1999 = -1919 (What?). The behavior is not normal.

goszowski:
It is default installation Sail + Octane from documentation and is not works.

PR #293, opened 2021-05-27
[1.x] Fix TokenMismatchException returns HTTP status code 200 instead of 419

### LINKED ISSUE #291: TokenMismatchException returns HTTP status code 200 instead of 419
- Octane Version: 1.0.2
- Laravel Version: 8.42.1
- PHP Version: 8.0.6

### Description:

Laravel's default exception handler sets the HTTP status code to 419 when a `Illuminate\Session\TokenMismatchException` is thrown by the system.

Running `php artisan serve` yields the expected result but not `php artisan octane:start`.

### Steps To Reproduce:

1. Create a new "test" route than only throws the above mentioned exception
2. Start the octane instance
3. Call the route from your browser of choice with network monitor enabled
4. Check that the responded status code is `200` instead of `419`

sy-records:
The status method if the second parameter is set, code can be any value, including the undefined HttpCode, such as 419.

fix #291

### DIFF
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -176,7 +176,7 @@ class SwooleClient implements Client, ServesStaticFiles
             }
         }
 
-        $swooleResponse->status($response->getStatusCode());
+        $swooleResponse->status($response->getStatusCode(), Response::$statusTexts[$response->getStatusCode()] ?? 'unknown status');
 
         foreach ($response->headers->getCookies() as $cookie) {
             $swooleResponse->{$cookie->isRaw() ? 'rawcookie' : 'cookie'}(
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Why is 419 undefined? That sounds like the real bug here?

sy-records:
419 is not a standard HTTP Code

GrahamCampbell:
It is an official status code, however: https://datatracker.ietf.org/doc/html/rfc6585#section-4.

sy-records:
this is 429, not 419

GrahamCampbell:
Oh. It could still be supported anyway. The status text should be `Authorization Timeout`.

sikhlana:
@GrahamCampbell I did add a PR that adds a reason for `419`: #294 
I also think it should return some kinda valid HTTP reason instead of `unknown status`.

taylorotwell:
Going to look at the other PR that adds the actual text.

PR #294, opened 2021-05-27
Fixes issue related to changing non-standard HTTP status codes to 200 OK

### LINKED ISSUE #291: TokenMismatchException returns HTTP status code 200 instead of 419

sikhlana:
This PR fixes #291 by passing an extra parameter `$reason` for non-standard HTTP codes.

### DIFF
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -19,6 +19,10 @@ use Throwable;
 
 class SwooleClient implements Client, ServesStaticFiles
 {
+    const STATUS_CODE_REASONS = [
+        419 => 'Page Expired',
+    ];
+
     public function __construct(protected int $chunkSize = 1048576)
     {
     }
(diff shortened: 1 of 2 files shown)


sikhlana:
Looks like I got ninja'd by @sy-records 😆

driesvints:
The other PR seems a bit more simpler so closing this one. Thanks!

chris-heng-lb:
Is it possible to let users define other status code reasons other than 419 through the `octane.php` config file?

LordVersA:
how can i add 481 or other statuses to this const?

PR #295, opened 2021-05-27
[1.x] Inject dependencies in Octane routes

cerbero90:
This PR allows Octane routes to have their dependencies injected:

```php
use Some\Bound\Service;
use Symfony\Component\HttpFoundation\Response;

Octane::route('GET', '/uri', fn (Service $service) => new Response($service->run()));
```

It also allows to define Octane routes without the leading `/`.

Usually I don't define routes with the leading slash so the first time I used Octane it took me some minutes to realize that Octane routes don't work if they don't begin with a slash. Of course this second feature is optional and can be discarded if you reckon it's an extra.

### DIFF
--- a/src/Concerns/ProvidesRouting.php
+++ b/src/Concerns/ProvidesRouting.php
@@ -3,7 +3,9 @@
 namespace Laravel\Octane\Concerns;
 
 use Closure;
+use Illuminate\Container\Container;
 use Illuminate\Http\Request;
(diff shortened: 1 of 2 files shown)


taylorotwell:
That slows the route down. The point is to be as minimal as possible. Service locate any dependencies you want inside the route. Or invoke your own action via the container, etc.

PR #296, opened 2021-05-27
[2.x] Let Octane routes call actions

cerbero90:
Based on https://github.com/laravel/octane/pull/295, this PR let Octane routes call actions similarly to the classic Laravel routes:

```php
Octane::route('GET', '/uri', 'Controller@action');
```
The supported actions are the same as classic routes: arrays, callables and strings.

This PR introduces breaking changes as it needs to slightly modify a couple of existing signatures.


taylorotwell:

PR #297, opened 2021-05-28
Display memory usage

stillzombie:
Since the Laravel documentation does not provide how to check for memory leaks, many developers will opt in to use the Laravel debug bar, but i found out that the debug bar it has a memory leak itself,  so displaying the memory usage besides the duration of each request will help to easy spot memory leaks without external tools.


driesvints:
This PR broke the test suite.

nunomaduro:
@driesvints On it.

stillzombie:
Sorry guys my bad, for the assertEquals on the test_request() function: 1 dot needs to be removed and also the display of the 500 request. the mb needs to be updated from '23.46' to '28.43'. I will be more careful while opening pr.

PR #313, opened 2021-06-09
Reset application scope to flush scoped instances

themsaid:
This PR calls `resetScope()` on the main container to reset any scoped instances.

It also moves uses `$sandbox->flush()` for tasks and ticks so the sandbox is cleared before removing it.


src/Listeners/FlushTemporaryContainerInstances.php
```diff
@@ -12,9 +12,12 @@ class FlushTemporaryContainerInstances
      */
     public function handle($event): void
     {
+        if (method_exists($event->app, 'resetScope')) {
+            $event->app->resetScope();
+        }
+
         foreach ($event->sandbox->make('config')->get('octane.flush', []) as $binding) {
             $event->app->forgetInstance($binding);
-            $event->sandbox->forgetInstance($binding);
```
taylorotwell, line 17: Would this be a breaking change?

themsaid: No, we already flush the entire sandbox after each operation.

PR #316, opened 2021-06-10
[1.x] Fix Swoole breaks when log is a valid JSON

### LINKED ISSUE #315: Logs formatted as valid json breaks octane
- Octane Version: 1.0.4
- Laravel Version: 8.46.0
- PHP Version: 8.0.7
- Database Driver & Version: N/A

### Description:

The swoole server acknowledges a log entry as something else (a request I guess) and routes it to `handleStream` when the log entry is a valid JSON.

I don't know if there is any other better way to solve it, but when receiving a json before routing it to `handleStream`, adding a check for the required `type` key as well at least solves the issue for the format in Monolog\Formatter\JsonFormatter

It would still conflict with other logs that:
1. Are valid json
2. Contains the key `type`

Will send a PR in a few minutes, please take a look

### Steps To Reproduce:

Create a laravel project and install octane

```bash
composer create-project --prefer-dist laravel/laravel laravel-octane
cd laravel-octane
composer require laravel/octane
php artisan octane:install
```

Add a log entry to AppServiceProvider so we don't need to call a route or anything just for demonstration purposes:
```php
public function boot()
{
    \Log::debug('a');
}
```

Run octane (that will call the log when booting the application):
```
APP_KEY=$(php artisan key:generate --show) LOG_CHANNEL=stderr LOG_STDERR_FORMATTER=Monolog\\Formatter\\JsonFormatter php artisan octane:start
```

henriquebremenkanp:
The swoole server acknowledges a log entry as something else (a request I guess) and routes it to handleStream when the log entry is a valid JSON.

I don't know if there is any other better way to solve it, but when receiving a json before routing it to handleStream, adding a check for the required type key as well at least solves the issue for the format in Monolog\Formatter\JsonFormatter

It would still conflict with other logs that:

1. Are valid json
2. Contains the key type

### DIFF
--- a/src/Commands/StartSwooleCommand.php
+++ b/src/Commands/StartSwooleCommand.php
@@ -177,7 +177,7 @@ class StartSwooleCommand extends Command implements SignalableCommandInterface
             ->filter()
             ->groupBy(fn ($output) => $output)
             ->each(function ($group) {
-                is_array($stream = json_decode($output = $group->first(), true))
+                is_array($stream = json_decode($output = $group->first(), true)) && isset($stream['type'])
                     ? $this->handleStream($stream)
                     : $this->error($output);
 


driesvints:
Please read the pr description before submitting and post a thorough explanation what this fixes. Don't just link to an issue. Thanks.

henriquebremenkanp:
@driesvints Sorry, just updated it

PR #317, opened 2021-06-10
[1.x] Fix Symfony uploaded file moving

tarampampam:
Related issue: https://github.com/spiral/roadrunner-laravel/issues/43
Same PR in https://github.com/spiral/roadrunner-laravel/pull/46


SelimSalihovic:
If you've started using Octane while in beta, upgrade it and make sure that your `octane.php` config file has the following entries under `listeners`:
```
WorkerStarting::class => [
    //whatever was there before
    EnsureUploadedFilesCanBeMoved::class, // import the class and add this line
],
```

Issue #320, opened 2021-06-11
Logging to stderr doesn't print anything


Radiergummi:
- Octane Version: `1.0.4`
- Laravel Version: `8.46.0`
- PHP Version: `8.0.5`
- Server & Version: ~Swoole~ / RoadRunner `2.2.1`
- Database Driver & Version: -

### Description:
When configuring the logging stack to write to `stderr` like the following:
```php
'stack' => [
    'driver' => 'stack',
    'channels' => ['stderr'],
    'ignore_exceptions' => false,
],
```
...any logged messages will not be printed if the app is started using `php artisan octane:start`. The roadrunner request log still prints properly, though.

For context, our application runs in a Docker stack where it's quintessential to have all log messages printed to standard output.

### Steps To Reproduce:
1. Create a new Laravel project:
    ```bash
    composer create-project laravel/laravel ./
    ```
2.  Add the octane dependency:
    ```bash
    composer require laravel/octane
    ```
3. Install octane:
    ```bash
   php artisan octane:install
    ```
    Make sure to select `0` for roadrunner and confirm all prompts.
4. Update the logging stack configuration to write to `stderr` instead of `single`:
    ```diff
    'channels' => [
        'stack' => [
            'driver' => 'stack',
    -       'channels' => ['single'],
    +       'channels' => ['stderr'],
            'ignore_exceptions' => false,
        ],
    ```
5. Update the index route handler to log a message using the `Log` facade:
    ```diff
    Route::get('/', function () {
    +   \Illuminate\Support\Facades\Log::error('test');
        return view('welcome');
    });
    ```
6. Start octane:
    ```bash
   php artisan octane:start
    ```
7. Open `http://localhost:8000` in the browser, monitor the log: No test message is logged.
8. For comparison, execute `php artisan serve`: A test message is logged

Radiergummi:
Reproduction repository: [Radiergummi/repro-laravel-octane-320](https://github.com/Radiergummi/repro-laravel-octane-320).

Radiergummi:
If logging to stderr requires additional steps, the documentation should include steps that detail them. Especially for Docker-based application stacks, stderr logging is best practice and should work out of the box with Laravel.

Namoshek:
This is because Roadrunner and Swoole spawn separate processes, which have their own process id. In containers, only the output of pid 1 is visible. In this case, pid 1 will be the `php artisan octane:start` process, which does not forward `stdout`/`stderr` of child processes.

[...]

PR #325, opened 2021-06-14
Added basic handling for application logs

Radiergummi:
This PR adds support for application logs written to stderr, as described in #320. As it stands, logging to stderr isn't possible in Octane, defeating best practices for Docker deployments.

For reference, here's my last post in that issue:
> For context on the issue itself, as Roadrunner uses stdout as its response transmission channel, it files all stderr output as application logs, with a level of `INFO`. This does not match any condition in the `getIncrementalOutput` line handler callback, and is discarded. So we can get application logs into the output by simply adding the following: [...]
>
> This has a few downsides, however:
> - **It prints noise messages:** Unfortunately, RR sends additional, non-app logs under the same channel ("INFO  scan command"). There's not really a good way to prevent this other than blacklisting them.
> - **It hard-codes INFO as the log level:** As we don't know which formatter the user has configured for the stderr logger, we can't simply parse the application-generated log line to extract the actual log level, and thus are forced to use INFO.
> - **It relies on an implementation detail:** This is of lesser concern, but (ab)using the `logger` field of the Roadrunner output for something it probably isn't intended for seems brittle to me.
>
> The first two points could be solved by adding special handling to Laravel itself: If it is running within Octane, log lines could be prefixed by some unique identifier, independently of the configured line formatter. That would allow checking for a substring match on the info lines received and only printing those.  
> Another option might be for Roadrunner itself to label application output differently, or adding a meta field to the log JSON.  
>
> Alternatively, we could simply add the three lines above and call it a day.

Anyone has ideas on how to go forward?

### DIFF
--- a/src/Commands/StartRoadRunnerCommand.php
+++ b/src/Commands/StartRoadRunnerCommand.php
@@ -182,6 +182,10 @@ class StartRoadRunnerCommand extends Command implements SignalableCommandInterfa
                         'duration' => $this->calculateElapsedTime($debug['elapsed']),
                     ]);
                 }
+
+                if ($debug['level'] === 'info' && $debug['logger'] === 'server') {
+                    return $this->info(trim($debug['msg']));
+                }
             });
 
         Str::of($server->getIncrementalErrorOutput())


driesvints:
@Radiergummi just want to remind you that PR's aren't reviewed until they're out of draft: https://laravel.com/docs/8.x/contributions#bug-reports

nunomaduro:
@Radiergummi Can you give me an example of a "logging" message that you think people would want to see and is currently missing on the octane output?

Namoshek:
Noteworthy @nunomaduro, the `stderr` logger works with Swoole already, only RR misses special handling.

Namoshek:
I'm having some trouble getting the `stderr` output to work. Would be nice to have the `stderr` logger actually log to `stderr`. I'm not sure if it is a WSL 2 issue though. Normally, within a command, `$this->getOutput()->getErrorStyle()` would return the `OutputStyle` for `stderr`. But for some reason it does not and the `stdout` one is returned instead (I see where the decision is made, but not why there is no `stderr` one assigned).

Anyway, the suggested solution does unfortunately not seem to work with the current existing configuration, because the `logs.level` parameter of RR is `debug` for `local` environments and `warning` for other environments (`production`, etc.). This will make `info` messages from RR vanish in production, removing all log output. Easy to test with `APP_ENV=production`.
    https://github.com/laravel/octane/blob/9ec7fa0052a56e46688165385ea02dfba0d43de7/src/Commands/StartRoadRunnerCommand.php#L87-L88

---

If we somehow get this to work, we should log the output without formatting for better readability and processability using `$this->getOutput()->writeln(trim($debug['msg']))` though. In my opinion it would also be nice if it was possible to change the formatting and/or disable the logging of requests (`200  GET  / .... 1ms`) as well - as part of another PR. Maybe this could even be done in the base `Command` class for all types of output.

src/Commands/StartRoadRunnerCommand.php
```diff
@@ -182,6 +182,10 @@ protected function writeServerOutput($server)
                         'duration' => $this->calculateElapsedTime($debug['elapsed']),
                     ]);
                 }
+
+                if ($debug['level'] === 'info' && $debug['logger'] === 'server') {
```
Namoshek, line 186: `$debug['level'] === 'info'` messages are not present in `APP_ENV=production`.

Namoshek (changes requested):
Just leaving a review to avoid accidental merging.

Namoshek:
The issue with `$debug['level'] === 'info'` seems fixable by adding `'-o', 'logs.channels.server.mode=production',` as RR parameter, which will override the log mode for the server channel only.

Namoshek:
Yes, using `--server=swoole` outputs the log messages as error messages:
```
ERROR  [2021-06-14 18:19:12] production.ERROR: test
```

The logs seem to be printed on `stdout` though, since the following command shows no output at all:
```sh
 php artisan octane:start --server=swoole >/dev/null
```

nunomaduro:
@Namoshek Alright - that is expected behavior? Or something seems odd there?

Namoshek:
Yeah, after thinking about it for another 2 minutes, I would say this cannot be intentional. Why should logs be printed as errors? Is it possible we are hitting the following code (by accident)?

https://github.com/laravel/octane/blob/9ec7fa0052a56e46688165385ea02dfba0d43de7/src/Commands/StartSwooleCommand.php#L179-L182

nunomaduro:
Currently investigating and working on this.

nunomaduro:
@Radiergummi @Namoshek Can you try this pull request https://github.com/laravel/octane/pull/326 on both drivers `swoole` and `roadrunner`, and let me know how it goes?

PR #326, opened 2021-06-15
Prints server logs to "stderr"

nunomaduro:
This pull request addresses https://github.com/laravel/octane/issues/320, https://github.com/laravel/octane/pull/325, by proxy applications logs (using `stderr` driver) to `stderr`.

Screenshot 2021-06-15 at 14 22 46

### DIFF
--- a/src/Commands/Concerns/InteractsWithIO.php
+++ b/src/Commands/Concerns/InteractsWithIO.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Octane\Commands\Concerns;
 
+use Illuminate\Console\OutputStyle;
 use Illuminate\Support\Str;
 use Laravel\Octane\Exceptions\DdException;
 use Laravel\Octane\Exceptions\ServerShutdownException;
@@ -19,9 +20,12 @@ trait InteractsWithIO
      *
      * @var array
      */
-    protected $ignoreErrors = [
+    protected $ignoreMessages = [
+        'scan command',
         'stop signal received, grace timeout is: ',
         'exit forced',
+        'worker constructed',
+        'worker destructed',
     ];
 
     /**
@@ -45,8 +49,21 @@ trait InteractsWithIO
      */
     public function error($string, $verbosity = null)
     {
-        if (! Str::contains($string, $this->ignoreErrors)) {
-            $this->label($string, $verbosity, 'ERROR', 'red', 'white');
+        $this->label($string, $verbosity, 'ERROR', 'red', 'white');
+    }
+
+    /**
+     * Write a string as raw output.
+     *
+     * @param  string  $string
+     * @return void
+     */
+    public function raw($string)
+    {
+        if (! Str::contains($string, $this->ignoreMessages)) {
+            $this->output instanceof OutputStyle
+                ? fwrite(STDERR, $string."\n")
+                : $this->output->writeln($string);
         }
     }
 
@@ -74,7 +91,7 @@ trait InteractsWithIO
      */
     public function label($string, $verbosity, $level, $background, $foreground)
     {
-        if (! empty($string)) {
+        if (! empty($string) && ! Str::startsWith($string, $this->ignoreMessages)) {
             $this->output->writeln([
                 '',
                 "  <bg=$background;fg=$foreground;options=bold> $level </> $string",
--- a/src/Commands/Concerns/InteractsWithIO.php
+++ b/src/Commands/Concerns/InteractsWithIO.php
@@ -60,7 +60,7 @@ trait InteractsWithIO
      */
     public function raw($string)
     {
-        if (! Str::contains($string, $this->ignoreMessages)) {
+        if (! Str::startsWith($string, $this->ignoreMessages)) {
             $this->output instanceof OutputStyle
                 ? fwrite(STDERR, $string."\n")
                 : $this->output->writeln($string);
--- a/src/Commands/StartRoadRunnerCommand.php
+++ b/src/Commands/StartRoadRunnerCommand.php
@@ -84,7 +84,7 @@ class StartRoadRunnerCommand extends Command implements SignalableCommandInterfa
             '-o', 'http.pool.supervisor.exec_ttl='.$this->maxExecutionTime(),
             '-o', 'http.static.dir=public',
             '-o', 'http.middleware=static',
-            '-o', app()->environment('local') ? 'logs.mode=production' : 'logs.mode=none',
+            '-o', 'logs.mode=production',
             '-o', app()->environment('local') ? 'logs.level=debug' : 'logs.level=warning',
             '-o', 'logs.output=stdout',
             '-o', 'logs.encoding=json',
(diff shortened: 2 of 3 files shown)


Radiergummi:
Can confirm this works well for Roadrunner, in both production and local environments. I'll test Swoole later today.

src/Commands/Concerns/InteractsWithIO.php
```diff
@@ -45,8 +49,21 @@ public function info($string, $verbosity = null)
...
+     *
+     * @param  string  $string
+     * @return void
+     */
+    public function raw($string)
+    {
+        if (! Str::startsWith($string, $this->ignoreMessages)) {
+            $this->output instanceof OutputStyle
+                ? fwrite(STDERR, $string."\n")
+                : $this->output->writeln($string);
```
Namoshek, line 66: I'm a bit confused by this piece of code - why does this work? Why check for `OutputStyle` to then not call a method on it? From what I can see, `$this->output` should always be `OutputStyle` within a Laravel `Command` anyway.

The call to `fwrite(STDERR, $str)` is ok for me, but seems like its not quite the Laravel way compared with the rest of the command base. I was fiddling around with it yesterday a bit and there should be some sort of `stderr` support in the Symfony base, but I couldn't get it to work properly. Is this also the reason why you have chosen to call `fwrite` on `stderr` directly?

src/Commands/StartRoadRunnerCommand.php
```diff
@@ -84,7 +84,7 @@ public function handle(ServerProcessInspector $inspector, ServerStateFile $serve
             '-o', 'http.pool.supervisor.exec_ttl='.$this->maxExecutionTime(),
             '-o', 'http.static.dir=public',
             '-o', 'http.middleware=static',
-            '-o', app()->environment('local') ? 'logs.mode=production' : 'logs.mode=none',
+            '-o', 'logs.mode=production',
```
Namoshek, line 87: This makes sense, but it would also be possible to just [override the logging mode for the `server` channel](https://roadrunner.dev/docs/intro-config) by adding:
```php
'-o', 'logs.channels.server.mode=production',
```
From what I can tell, this would suppress other noisy messages in production. Interestingly though, the mode `none` is not even documented...

Namoshek (commented):
I only have a very simple application to test this changes right now, so I cannot really go into feature test details or anything.

---

The solution does what the `stderr` logger suggests - it logs server messages on `stderr`. It does this for all server messages though, no matter if they come from `stdout` or `stderr`. It also does not respect the actual logger config of the application. I'm not sure if it should though (because I've no idea what stuff RR and swoole print besides the application logs).

An ideal solution, in my opinion, would only forward `stderr` output of the worker processes to `stderr` of the main process. When reading the RR documentation, I though this could be achieved with `logs.err_output` setting, but it does not seem to do what the name suggests. So maybe this isn't even possible.

src/Commands/StartRoadRunnerCommand.php
nunomaduro: Let's keep the "logs.mode=production" for now.

src/Commands/Concerns/InteractsWithIO.php
nunomaduro: If the condition `$this->output instanceof OutputStyle` passes, this means we are the CLI context, therefore we should/can use the `fwrite` safely. Symfony does support writing for `stderr` using the `$output->getErrorOutput()`, however the `$output` instance we are using at Laravel does not have support for that.

nunomaduro:
@Namoshek Your last comment was not clear for me. At this time, **only** server messages - outputted by the Laravel application - are captured and proxy to the `stderr` by Octane.

src/Commands/Concerns/InteractsWithIO.php
Namoshek: Ok got it. Doesn't make much sense to run the `octane:start` command from within php-fpm for example, but I realize its possible. This part seems fine then!

Namoshek:
@nunomaduro Exactly, everything from the `server` channel. What I was aiming at: for me is unclear whether the `server` channel contains only application logs (from the `stderr` logger) or also other logs/messages which I could not trigger in my very basic setup. If the latter is the case, we probably need some additional checks besides the ignored messages you already added. But I'm fine with adding those as they come up.

The only thing which is still a bit weird to me: why are `stderr` logger logs not on the `stderr` output of the worker process? Only forwarding `stderr -> stderr` and not `stdout -> stderr` would be the best solution after all, or am I missing something?

nunomaduro:
> @nunomaduro Exactly, everything from the server channel. What I was aiming at: for me is unclear whether the server channel contains only application logs (from the stderr logger) or also other logs/messages which I could not trigger in my very basic setup. If the latter is the case, we probably need some additional checks besides the ignored messages you already added. But I'm fine with adding those as they come up.

Agree.

> The only thing which is still a bit weird to me: why are stderr logger logs not on the stderr output of the worker process? Only forwarding stderr -> stderr and not stdout -> stderr would be the best solution after all, or am I missing something?

That seems to be a roadrunner thing. We actually don't even touch on that configuration, so I would also expect those to be in the `stderr`. 

Screenshot 2021-06-15 at 19 30 14

Radiergummi:
> An ideal solution, in my opinion, would only forward stderr output of the worker processes to stderr of the main process. When reading the RR documentation, I though this could be achieved with logs.err_output setting, but it does not seem to do what the name suggests. So maybe this isn't even possible.

As far as I know from a Spiral project, the `err_output` actually refers to error messages emitted by the roadrunner process, but it doesn't relate to the PHP application hosted by Roadrunner. It uses stdout to emit HTTP responses, so the only side channel left for logs is stderr.

> @Radiergummi Please try to also test in some edge cases, to ensure this change won't effect with users.

I monkey-patched our main production workload with the PR changes, compared the output to our php-fpm variant, and did not notice any differences in log output. Printing huge messages (`str_repeat('x', 9_000_000)`) worked fine, as did stack traces from the exception handler including line breaks. As far as I can see, this feature works as intended and restores the ability to write logs in roadrunner deployments.

Concerning the server channel: It would be preferable if Roadrunner had a log channel exclusive to the hosted application, or alternatively sent unrelated messages to another channel. Maybe they'd be open for such a change?

nunomaduro:
> Concerning the server channel: It would be preferable if Roadrunner had a log channel exclusive to the hosted application, or alternatively sent unrelated messages to another channel. Maybe they'd be open for such a change?

Not sure - can you ask them? Meanwhile, making this PR open for review.

Radiergummi:
> Not sure - can you ask them? Meanwhile, making this PR open for review.

Yep, I planned to open a Roadrunner issue on this. But if that is implemented at all, it's going to take a while, so it's out of scope for this PR anyway.

Radiergummi:
@nunomaduro is there an ETA on the next release? I'm just now trying to debug an issue in production, but I don't have any stack traces in the application logs...

driesvints:
@Radiergummi releases are done on Tuesday's.

PR #327, opened 2021-06-18
Set roadrunner http.middlware as an option

foxx9:
Roadrunner `http.middlware` config is currently hardcoded to "static" in  `src/Commands/StartRoadRunnerCommand.php` file, and thus cannot be changed via .rr.yaml file

This pull request set this as an option in `config/octane` file in order to customize it following the official conf : https://roadrunner.dev/docs/intro-config

The main purpose is to allow to enable gzip as this is one the supported middleware option.
Ex : `    'http_middleware' => 'static,gzip'`

### DIFF
--- a/config/octane.php
+++ b/config/octane.php
@@ -136,20 +136,6 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | RoadRunner middleware option
-    |--------------------------------------------------------------------------
-    |
-    | While using Roadrunner, you may change the http middleware setting.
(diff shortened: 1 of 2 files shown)


foxx9:
Hi @taylorotwell , Thanks for accepting the request,  however I see that you have used 'swoole' in the config file path in you commit:  config('octane.swoole.http_middleware', 'static')

Did you mean 'octane.roadrunner.http_middleware' instead ? 

thanks

Issue #332, opened 2021-06-22
Change path to config file


vpatrikeev:
It would be nice to have opportunity to switch config file or add output to it as octane:roadrunner command does. For example, we have different configs for dev and prod and now we have to copy contents to .rr.yaml depending on env

driesvints:
We don't have plans to work on this ourselves but if you could attempt a PR so we can look at actual code, that would be great.

PR #335, opened 2021-06-28
Addresses #332: pass config-file path for roadrunner

kevincobain2000:
Submitted with intentions for a first review and consideration.

Reasons for this pull request:

1. As mentioned in #332: to be able use different ``.rr.yaml`` based on different environments
2. We use ``deployer``. As this is symbolic link deployment, the ``base_path()`` picks up the path from the ``releases/<id>`` reference instead of the symb link. It'd be nice if we could provide a path via cli.

The first option is more encouraging, as this config file is most likely to be on the server in a different location.

### DIFF
--- a/src/Commands/StartRoadRunnerCommand.php
+++ b/src/Commands/StartRoadRunnerCommand.php
@@ -22,6 +22,7 @@ class StartRoadRunnerCommand extends Command implements SignalableCommandInterfa
      */
     public $signature = 'octane:roadrunner
                     {--host=127.0.0.1 : The IP address the server should bind to}
+                    {--config-path= : The config path for roadrunner .rr.yaml file}
                     {--port=8000 : The port the server should be available on}
                     {--rpc-port= : The RPC port the server should be available on}
                     {--workers=auto : The number of workers that should be available to handle requests}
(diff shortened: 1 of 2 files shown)


src/Commands/StartRoadRunnerCommand.php
```diff
@@ -120,6 +119,20 @@ protected function writeServerStateFile(
...
+     *
+     * @return string
+     */
+    protected function configPath()
+    {
+        if (! $this->option('config-path')) {
+            touch(base_path('.rr.yaml'));
+            return base_path('.rr.yaml');
+        }
+        return $this->option('config-path');
```
Namoshek, line 133: Would it make sense here to allow different styles of config paths to be passed?
```
php artisan octane:start --config-path=rr.prod.yml
php artisan octane:start --config-path=config/rr.prod.yml
php artisan octane:start --config-path=./config/rr.prod.yml
php artisan octane:start --config-path=../config/rr.prod.yml
php artisan octane:start --config-path=/etc/roadrunner/rr.prod.yml
```

kevincobain2000, line 133: Something like below you mean?

```suggestion
        $configPath =  $this->option('config-path');
        if (! $this->isAbsolutePath($configPath)) {
            return base_path($configPath);
        }
        return $configPath;
```

And have a new function isAbsolutePath()

```
>>> function isAbsolutePath($path) {
...     if($path === null || $path === '') throw new Exception("Empty path");
...     return $path[0] === DIRECTORY_SEPARATOR || preg_match('~\A[A-Z]:(?![^/\\\\])~i',$path) > 0;
... }
```

src/Commands/StartRoadRunnerCommand.php
```diff
@@ -120,6 +119,20 @@ protected function writeServerStateFile(
...
+    }
+
```
kevincobain2000, line 135:
```suggestion

private function isAbsolutePath($path) {
    if($path === null || $path === '') throw new Exception("Empty path");
    return $path[0] === DIRECTORY_SEPARATOR || preg_match('~\A[A-Z]:(?![^/\\\\])~i',$path) > 0;
}
```

src/Commands/StartRoadRunnerCommand.php
kevincobain2000: Thank you, I have added absolute path 58f1dcf

driesvints:
The flag was renamed to `--rr-config`.

PR #355, opened 2021-08-05
[1.x] Resolve components when needed

foremtehan:
In my app i dont have any Auth, Cookie, Session system (telegram app) so:


driesvints:
@foremtehan tests seem to be failing.

PR #358, opened 2021-08-06
Fix typo

sy-records:
Install the Swoole dependencies.

https://github.com/laravel/octane/blob/7de906ad095741819f4c2d16741620d1d1bc2caf/src/Commands/InstallCommand.php#L104-L118


driesvints:
Thank you!

PR #385, opened 2021-09-25
Add clear the cache of APCu and Opcache

### LINKED ISSUE #384: octane:reload with opcache_reset()
Hi,

This is a feature request together with a question.

**Question** When using ``swoole`` or ``rr``, **is it recommended to** **_also_** use ``opcache``?
It is not clear to me yet which is the suggested way. As per my tests, app on ``swoole`` combined with ``opcache`` have better performance and consumes less memory per request, than just ``swoole``.

**Problem**

If the answer to question is **Yes** 
As with ``swoole`` or ``rr``, php cli is used with ``opcache.enable_cli=1``. 

1. **Where is the best place to do ``opcache_reset();``** for all the workers at the time of deployment (``php artisan octane:reload``)? 

If this a feature that can be provided by octane, as ``opcache_reset`` needs to be done for all the workers, and currently, there is only one signal ``SIGUSR1`` that is sent to the master process id and swoole-server independently reloads the servers from there on.

sy-records:
Fix #384

If PHP has APCu/OPcache enabled, reloads will be affected when reloading, so you need to clear the cache

### DIFF
--- a/src/Swoole/Handlers/OnWorkerStart.php
+++ b/src/Swoole/Handlers/OnWorkerStart.php
@@ -37,6 +37,7 @@ class OnWorkerStart
 
         $this->dispatchServerTickTaskEverySecond($server);
         $this->streamRequestsToConsole($server);
+        $this->clearCache();
 
         if ($this->shouldSetProcessName) {
             $isTaskWorker = $workerId >= $server->setting['worker_num'];
@@ -108,5 +109,26 @@ class OnWorkerStart
                 (microtime(true) - $this->workerState->lastRequestTime) * 1000,
             );
         });
+    }
+
+    /**
+     * Clear the cache of APCu and Opcache.
+     *
+     * @return void
+     */
+    protected function clearCache()
+    {
+        static $functions = [
+            'apcu_clear_cache',
+            'opcache_reset',
+        ];
+
+        foreach ($functions as $function)
+        {
+            if (function_exists($function))
+            {
+                $function();
+            }
+        }
     }
 }
--- a/src/Swoole/Handlers/OnWorkerStart.php
+++ b/src/Swoole/Handlers/OnWorkerStart.php
@@ -123,10 +123,8 @@ class OnWorkerStart
             'opcache_reset',
         ];
 
-        foreach ($functions as $function)
-        {
-            if (function_exists($function))
-            {
+        foreach ($functions as $function) {
+            if (function_exists($function)) {
                 $function();
             }
         }


taylorotwell:
What does this actually fix? The OP of the issue doesn't actually talk about what their problem is. How can I recreate a problem that this code change fixes?

sy-records:
When cache extensions such as `OPcache ` are enabled, using `reload` or `--watch` will cause the new content to not take effect. @taylorotwell

sy-records:
@kevincobain2000 `--max-requests` will also pull up the process again, and both will call the onWorkerStart callback, so maybe we can add an environment judgment

sy-records:
> Environment judgement won't work -  as the env at the time of `octane:start` on the original worker won't be overridden by new env value at the time of `octane:reload`.

It won't overwrite is right, it's reload, not restart

hughcube:
Why clear opCache when each worker starts?    This should be decided by the user (via `php.ini` or `php -d opcache.enable_cli=0`)
There is no way to use the optimization brought by opCache every time it is cleared, especially in the scenario of `--max-requests=1`

src/Swoole/Handlers/OnWorkerStart.php
```diff
@@ -37,6 +37,7 @@ public function __invoke($server, int $workerId)
 
         $this->dispatchServerTickTaskEverySecond($server);
         $this->streamRequestsToConsole($server);
+        $this->clearCache();
```
huangdijia, line 40: 每个 worker 在 start 的时候都跑一遍？

hughcube: 是这个意思的,   你的  `--max-requests`  设置的越大影响越小

kevincobain2000:
@hughcube 

> Why clear opCache when each worker starts? 

Each worker starts upon 1) reloading application or 2) when ``--max-requests`` have reached. Ideally, it is better to clear cache when app is reloaded, however there is no way to judge that. 

> 是这个意思的, 你的 --max-requests 设置的越大影响越小

Therefore, yes

hughcube:
@kevincobain2000 
It’s not the point, I think it should be controlled by the user himself
`php.ini` or `php -d opcache.enable_cli=0`

kevincobain2000:
@hughcube 

I am not sure if I get what you mean by 
>  I think it should be controlled by the user himself
> php.ini or php -d opcache.enable_cli=0

You mean control enable/disable of opcache or flush opcache? 
Or you mean, Ppcache is not necessary anymore since it is already cached by swoole or rr? I also want to know this that do we really need Opcache and does it add value anymore?

For the apps that have opcache enabled, flushing opcache cannot be controlled by user, unless the Octane app is restarted. Unlike fpm, where fpm:reload flushes opcache, cli cannot flush opcache. Therefore, in the case of octane:reload (in case of new deployment), cache codes need to be flushed.

hughcube:
@kevincobain2000 


I understand what you mean. 

We now use docker deployment. When building, the opcode of commonly used php files will be cached first, and set `--max-requests=1`, so clearing the cache every time I start the worker will increase my overhead. 

Clear the cache when octane starts, such as this
`php artisan octane:start --clear-opcode-cache=1`

Clearing the cache when the worker starts can be used for hot updates, such as this
`php artisan octane:start --clear-opcode-cache=2`

I think clearing the cache is an option, or even the default, but it is not necessary.

kevincobain2000:
@hughcube 
I agree that clearing the cache code should be optional, especially for the use cases like yours as you are using docker and octane:start is going to flush the cache anyways.

PR #400, opened 2021-10-10
Add Livewire listener

joshhanley:
This adds a listener for Livewire to ensure all state is flushed before each operation. 

A corresponding PR https://github.com/livewire/livewire/pull/3987 has been submitted to Livewire.

Hope this helps!


calebporzio:
Seconding the need for this. Thanks!

woodspire:
Instead of adding a listener for all the external packages that would need to be flushed, wouldn't be better to generate an event (FlushRequest) in octane that livewire would register and flush itself ? This way all external packages can be flushed without a PR to octane each time.

PR #407, opened 2021-10-20
Since 4.1.7, swoole is called openswoole

XternalSoft:
Since 4.1.7, swoole is called openswoole


sy-records:
openswoole is a forked, not a replacement

PR #420, opened 2021-11-15
[1.x] Adds tests against "openswoole"

nunomaduro:
This pull request duplicates the tests workflow, so we can test octane against "openswoole". A good news, seems that "open-swoole" is already compatible with PHP 8.1. Yet, "swoole" is not, so we are going to ignore the PHP 8.1 swoole build for now.


driesvints:
@nunomaduro I'd personally keep this grouped under a single file as a different "job". The way we usually organise GHA workflows is to group specific "type" of workflows. 

Like framework: types, tests, database engines.

Swoole and OpenSwoole tests basically test the actual working of Octane so it's best to keep these grouped under "tests.yml". All OS projects of Laravel also have a "tests.yml" file.

nunomaduro:
Checking Dries...

.github/workflows/tests.yml
```diff
@@ -7,7 +7,7 @@ on:
     - cron: '0 0 * * *'
 
 jobs:
-  tests:
+  tests / swoole:
```
driesvints, line 10:
```suggestion
  swoole:
```

.github/workflows/tests.yml
```diff
@@ -37,3 +37,34 @@ jobs:
 
       - name: Execute tests
         run: vendor/bin/phpunit --verbose
+        continue-on-error: ${{ matrix.php > 8 }}
+  tests / openswoole:
```
driesvints:
```suggestion
  openswoole:
```

Issue #426, opened 2021-11-26
Process->getIncrementalOutput() does not flush output buffer


tezhm:
- Octane Version: 1.0.18
- Laravel Version: 8.70.2
- PHP Version: 8.0.13
- Server & Version: Swoole / RoadRunner 2.5.6
- Database Driver & Version:

### Description:

`StartRoadRunnerCommand->writeServerOutput()` queries the worker output via `Process->getIncrementalOutput()` which internally buffers the worker process output. This buffer is not flushed and thus keeps growing indefinitely.

I have worked around this by calling `Process->clearOutput()` and `Process->clearErrorOutput()` after each call to `StartRoadRunnerCommand->writeServerOutput()`.

### Steps To Reproduce:

Start octane server for roadrunner binary and log to stderr from workers.

nunomaduro:
@tezhm What exactly makes you think that "This buffer is not flushed and thus keeps growing indefinitely."? What have you used to conclude this?

tezhm:
Process stores output logs to php://temp. This buffer temporarily stores
logs to file after output reaches a certain size. The
`getIncrementalOutput()` iterates over output rather than flushing it.
Reason for this is so all output can be queried later using `getOutput()`.
Thus, if output is never flushed the temporary file will grow indefinitely.

PR #428, opened 2021-11-28
[1.x] Fixes Octane's process output not being flushed

nunomaduro:
This pull request fixes an edge case where the temporary file created by Octane's Symfony process for the process's output, may take a considerable disk space, if the process is running for too long. In addition, we don't really need that full output, as we already use it the `getIncrementalOutput` for getting incrementally the output.

So, the solution in this pull request is simple, we just clear the output - the temporary file - every time we grab information from the output.

Closes https://github.com/laravel/octane/issues/426

### DIFF
--- a/src/Commands/Concerns/InteractsWithServers.php
+++ b/src/Commands/Concerns/InteractsWithServers.php
@@ -107,7 +107,7 @@ trait InteractsWithServers
     }
 
     /**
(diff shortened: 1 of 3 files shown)


nunomaduro:
Note, tests will pass once this gets merged: https://github.com/laravel/octane/pull/429.

PR #437, opened 2021-12-13
[1.x] Add new option that strict raw server output for swoole.

chris-lee-lb:
This pull request addresses #341 for the swoole part. And this introduce the new option - `raw-server-output` to make sure we only use raw format for method `writeServerOutput` in `StartSwooleCommand`

### DIFF
--- a/src/Commands/StartCommand.php
+++ b/src/Commands/StartCommand.php
@@ -22,7 +22,8 @@ class StartCommand extends Command implements SignalableCommandInterface
                     {--task-workers=auto : The number of task workers that should be available to handle tasks}
                     {--max-requests=500 : The number of requests to process before reloading the server}
                     {--rr-config= : The path to the RoadRunner .rr.yaml file}
-                    {--watch : Automatically reload the server when the application is modified}';
+                    {--watch : Automatically reload the server when the application is modified}
+                    {--raw-server-output : Only use raw format for server output}';
 
     /**
      * The command's description.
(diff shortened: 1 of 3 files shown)


nunomaduro:
@chris-lee-lb What is the use case?

chris-lee-lb:
@nunomaduro In modern docker infrastructure, we usually has this log streams - `applications -> os standard output -> docker log drivers -> AWS CloudWatch or GCP StackDriver logging`

And AWS CloudWatch / GCP StackDriver logging all support parse json string in object, for advanced log searching. (<https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/CWL_AnalyzeLogData-discoverable-fields.html>)

If we have such prefix `INFO`, it will not be treat as valid json.

nunomaduro:
I feel that we should manage this internally checking the Symfony's Command Verbosity Levels, instead of being a new option. Let's see what Taylor thinks about it.

taylorotwell:
@nunomaduro can you elaborate? using your approach how would one indicate they want raw output?

Namoshek:
> I feel that we should manage this internally checking the Symfony's Command Verbosity Levels, instead of being a new option. Let's see what Taylor thinks about it.

I don't think the verbosity **level** should be used to select the output **format**. I'm not sure `--raw-server-output` is the ideal name for the option either, but it is more clear in my opinion. Although if it was up to me, I would probably suggest adding a `--log-format=default` option with `--log-format=json` as variant.

taylorotwell:
I do agree I don't think a log verbosity setting should change the log format.

chris-lee-lb:
@nunomaduro @Namoshek @taylorotwell thanks for all your feedback.

So I will change  option name to `--log-format`, with two allowed values

* default 
* raw

Not consider `json` in the feedback was because it may too complex to handle at framework level, but `raw` is more simple to implement and also is suitable for original purpose.

chris-lee-lb:
@nunomaduro all changes are done. Thanks you.

nunomaduro:
Going to re-work this pull request in a different branch.

PR #438, opened 2021-12-14
[1.x] Adds `json` log formatter

nunomaduro:
This pull request **is a work in progress**, and it adds a `json` log formatter instead of the original `raw` proposed one. It also simplifies the original pull request a little bit.

@Namoshek , @chris-lee-lb can you let me know if this satisfies the original needs you had? If yes, I will polish this pull request, add tests, etc.

Screenshot 2021-12-14 at 11 08 03
.

### DIFF
--- a/src/Commands/Concerns/InteractsWithIO.php
+++ b/src/Commands/Concerns/InteractsWithIO.php
@@ -3,6 +3,7 @@
 namespace Laravel\Octane\Commands\Concerns;
 
 use Illuminate\Support\Str;
+use Laravel\Octane\Commands\Command;
 use Laravel\Octane\Exceptions\DdException;
 use Laravel\Octane\Exceptions\ServerShutdownException;
 use Laravel\Octane\Exceptions\WorkerException;
@@ -119,7 +120,7 @@ trait InteractsWithIO
         $duration = number_format(round($request['duration'], 2), 2, '.', '');
 
         $memory = isset($request['memory'])
-            ? number_format($request['memory'] / 1024 / 1204, 2, '.', '')
+            ? (number_format($request['memory'] / 1024 / 1204, 2, '.', '').' mb ')
             : '';
 
         ['method' => $method, 'statusCode' => $statusCode] = $request;
@@ -153,7 +154,7 @@ trait InteractsWithIO
                $method,
                $url,
                $dots,
-               empty($memory) ? $memory : ($memory.' mb '),
+               $memory,
                $duration,
             ), $this->parseVerbosity($verbosity)),
         };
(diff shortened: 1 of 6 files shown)


src/Commands/Concerns/InteractsWithIO.php
```diff
@@ -29,17 +28,15 @@ trait InteractsWithIO
...
     /**
-     * Write a string as raw output.
+     * Gets the log format, if any.
      *
-     * @param  string  $string
-     * @return void
+     * @return string|null
      */
-    public function raw($string)
+    public function logFormat()
```
driesvints, line 35: I don't think we can remove a method on a stable release here?

nunomaduro: I can add it back - but honestly, this is internal API, people should not really try extend octane commands...

driesvints: @nunomaduro I also very much doubt anyone's doing that but I've seen stranger things happening in the past 😅

Namoshek: Since this PR does not remove the calls to `raw` in `StartSwooleCommand` for example, this for sure cannot be safe or am I missing something?

Namoshek (commented):
This looks mostly what I had in mind, yes. Some thoughts though:

- Currently, an exception log entry is hard to differentiate from a request log entry. Would it make sense to introduce a `type=request|exception|info` for each entry?

- Extending the built-in supported log formats (`default` and `json`) is almost impossible with this solution (not so much because of `logFormat()` but the `match` statements). I am by no means a logging expert and I also do not have concrete plans to do so, but I think there are some standard log formats which one might want to implement in user-land code or a third-party packages to avoid having to transform logs outside the application.

- From what I can tell, the selected `--log-format` also impacts the output channel, i.e. `stdout` or `stderr`. Is this expected behavior from a user perspective?

src/Commands/Concerns/InteractsWithIO.php
GrahamCampbell: `@internal` tag if your friend. ;)

chris-lee-lb:
@nunomaduro And for the part of RoadRunner , it seems that it has certain output formats if you want to write logs into stdout. otherwise you will keep receive this error message and  RoadRunner worker will be crashed.

```
error: sync_worker_exec_worker_with_timeout: Network:
	sync_worker_exec_payload:
	goridge_frame_receive: CRC verification failed, bad header: {"type":"audit-log","severity":"INFO","times, worker's pid: 28
worker stopped, error: signal: killed; process_wait: signal: killed
worker allocated, pid: 35
```

So may not apply this feature to `src/Commands/StartRoadRunnerCommand.php` is more ideal for now.

PS: I do not use custom rr-config file.

chris-lee-lb:
@nunomaduro after more research, I've founded out that I can always get raw output if I write logs into `STDERR` for both Swoole and RoadRunner.

So for my original purpose, it already can be done. Sorry for the bother. Maybe we can update the documentation to note this situation.

PR #439, opened 2021-12-16
Update swoole's website

sorshion:

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell (changes requested):
That is the wrong swoole. We are using openswoole, not the original.

GrahamCampbell:
https://github.com/openswoole/swoole-src

driesvints:
@GrahamCampbell that's not true. We're offering support for both.

nunomaduro:
I think we should do https://github.com/laravel/octane/pull/441, in favour of this pull request.

driesvints:
@nunomaduro yeah I agree

PR #502, opened 2022-04-07
Make the bin files used by Swoole and Roadrunner config options

jedjones-uk:
Hi,

This is a simple but very helpful change, instead of the bin files used to run the servers being hard coded, they are made configurable.

The reason for doing so is without this developers are forced to extend first the server specific command then the octane:start command if they want to change anything about the boot/worker spawn process.

This is particularly useful for roadrunner if you want to support having roadrunner itself act as a queue consumer as that requires a different setup to PSR7Worker.

The config is defaulted to identical values it had before.

I think the escape hatch it provides is powerful and the increase in maintenance burden is negligible.

### DIFF
--- a/src/Commands/StartRoadRunnerCommand.php
+++ b/src/Commands/StartRoadRunnerCommand.php
@@ -79,7 +79,7 @@ class StartRoadRunnerCommand extends Command implements SignalableCommandInterfa
             '-c', $this->configPath(),
             '-o', 'version=2.7',
             '-o', 'http.address='.$this->option('host').':'.$this->option('port'),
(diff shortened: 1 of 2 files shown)


taylorotwell:
Why is one of them `{driver}.command` and the other is `{driver}.server.command`?

jedjones-uk:
I went back and forth over using something like `{driver}.bin`/`{driver}.command` but ultimately decided to match the roadrunner option to the path it uses to override the config with cli properties.

It didn't look like Swoole had a similar functionality hence just going with root array.

I'm not married to the choice either way, just went with what made sense at the time.

taylorotwell:
Thanks!

PR #506, opened 2022-04-11
Force the processId to int

sy-records:
If the `$this->serverStateFile->read()` file is not readable, `ProcessId` will return null

Maybe we can change it to 0


huangdijia:
只开放组织内成员提 pr?

sy-records:
GitHub doesn't allow maintainer permissions from organization accounts.

So we need to use our personal account to submit PR.

image

PR #507, opened 2022-04-11
Force the processId to int

sy-records:
If the `$this->serverStateFile->read()` file is not readable, `ProcessId` will return null

Maybe we can change it to 0


driesvints (approved):

driesvints:
Thanks @sy-records

PR #510, opened 2022-04-18
Fix max_execution_time config doesn't work

### LINKED ISSUE #470: `max_execution_time` config doesn't work
- Octane Version: 1.1.1
- Laravel Version: 8.77.1
- PHP Version: 8.1.0
- Server & Version: Swoole 4.8.3

### Description:
When the execution time for requests being handled by Octane exceeded from `max_execution_time` in `octane.php` config file, the Octane wouldn't send any response and the request remains in the pending state.

### Steps To Reproduce:

Set `max_execution_time` to `2` seconds in `octane.php`:

```php
// config/octane.php
return [
    'max_execution_time' => 2,
];
```

and try:
```php
Route::get('/foo', function() {
    sleep(3);
    return 'bar';
});
```

I know this may be a Swoole bug. So if confirmed, I'll open an issue on the Swoole repo.

sy-records:
Fix #470

Re-create the new Swoole response object to use it to send the response.

image

### DIFF
--- a/bin/createSwooleTimerTable.php
+++ b/bin/createSwooleTimerTable.php
@@ -10,6 +10,7 @@ if (($serverState['octaneConfig']['max_execution_time'] ?? 0) > 0) {
 
     $timerTable->column('worker_pid', Table::TYPE_INT);
     $timerTable->column('time', Table::TYPE_INT);
+    $timerTable->column('fd', Table::TYPE_INT);
 
     $timerTable->create();
 
--- a/src/Swoole/Actions/EnsureRequestsDontExceedMaxExecutionTime.php
+++ b/src/Swoole/Actions/EnsureRequestsDontExceedMaxExecutionTime.php
@@ -3,13 +3,16 @@
 namespace Laravel\Octane\Swoole\Actions;
 
 use Laravel\Octane\Swoole\SwooleExtension;
+use Swoole\Http\Response;
+use Swoole\Http\Server;
 
 class EnsureRequestsDontExceedMaxExecutionTime
 {
     public function __construct(
         protected SwooleExtension $extension,
         protected $timerTable,
-        protected $maxExecutionTime
+        protected $maxExecutionTime,
+        protected ?Server $server = null
     ) {
     }
 
@@ -25,6 +28,12 @@ class EnsureRequestsDontExceedMaxExecutionTime
                 $this->timerTable->del($workerId);
 
                 $this->extension->dispatchProcessSignal($row['worker_pid'], SIGKILL);
+
+                if ($this->server instanceof Server && isset($row['fd'])) {
+                    $response = Response::create($this->server, $row['fd']);
+                    $response->status(408);
+                    $response->end();
+                }
             }
         }
     }
--- a/src/Swoole/Actions/EnsureRequestsDontExceedMaxExecutionTime.php
+++ b/src/Swoole/Actions/EnsureRequestsDontExceedMaxExecutionTime.php
@@ -29,7 +29,7 @@ class EnsureRequestsDontExceedMaxExecutionTime
 
                 $this->extension->dispatchProcessSignal($row['worker_pid'], SIGKILL);
 
-                if (extension_loaded('swoole') && $this->server instanceof Server) {
+                if ($this->server instanceof Server) {
                     $response = Response::create($this->server, $row['fd']);
                     $response->status(408);
                     $response->end();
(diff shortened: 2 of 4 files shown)


bin/createSwooleTimerTable.php
```diff
@@ -10,6 +10,7 @@
 
     $timerTable->column('worker_pid', Table::TYPE_INT);
     $timerTable->column('time', Table::TYPE_INT);
+    $timerTable->column('fd', Table::TYPE_INT);
```
driesvints, line 13: Isn't this a breaking change? What will happen for existing Octane installations that don't have this column?

src/Swoole/Actions/EnsureRequestsDontExceedMaxExecutionTime.php
```diff
@@ -3,13 +3,15 @@
...
+use Swoole\Http\Response;
 
 class EnsureRequestsDontExceedMaxExecutionTime
 {
     public function __construct(
         protected SwooleExtension $extension,
         protected $timerTable,
-        protected $maxExecutionTime
+        protected $maxExecutionTime,
+        protected $server
```
driesvints, line 14: Even though unlikely, if this action is used directly by users, it'll result in a breaking change.

bin/createSwooleTimerTable.php
sy-records: It should not be BC, it will be overwritten after updating the version.

driesvints: You mean it should be BC? Because if it isn't then we need to target this PR to master.

sy-records: It's not. 
It is an enhancement.

driesvints: I feel like there's a slight language barrier here. I'm trying to determine if apps using Octane NOW will break if we ship this. Can you confirm they won't break saying "they won't break"?

sy-records: If a new version is released after the merger, the bin directory will be updated with the version. Will this cause any problems?

driesvints: That's what I'm asking you 😅 

You're setting a value in this table here: https://github.com/laravel/octane/pull/510/files#diff-d72d084dbc08724cd72e4665b6da8495735c3f21b16e2ea8c6ac669fb56f3b0cR110

But existing Octane apps on the 1.x release won't have this column. I have no idea what Swoole will do if that happens. That's why I want to make sure that nothing breaks for people who don't have this column.

Namoshek: From my understanding, Swoole only offers the generic table framework and the discussed change is Octane specific. And since Swoole tables are in-memory only, this should be BC as long as Octane is restarted after an update.

driesvints: @Namoshek thanks for that. Since we recommend that in the docs, I feel that this change is okay 👍 

https://laravel.com/docs/9.x/octane#reloading-the-workers

taylorotwell:
What is `fd`? Can you explain more what this PR is actually doing?

sy-records:
The client connection $fd number

you can see https://openswoole.com/docs/modules/swoole-http-response-create

taylorotwell:
This is available on all versions of Swoole? Both Swoole and OpenSwoole?

sy-records:
swoole is supported since v4.6.0+, openswoole is forked from v4.7.0, so both are supported.

But openswoole has a [max_request_execution_time](https://openswoole.com/docs/modules/swoole-server/configuration#max_request_execution_time) configuration, so I didn't add it for it.

smortexa:
@sy-records I think OpenSwoole should be added because `max_request_execution_time` config doesn't work on Octane yet (openswoole/swoole-src#136).

sy-records:
Got it, thanks for the feedback! @smortexa

taylorotwell:
Sorry - still trying to understand this. How does this fix the problem with a worker hanging? Isn't this response being created on the master process?

sy-records:
Since the worker process is forced to kill, it is understood to have been [detached](https://openswoole.com/docs/modules/swoole-http-response-detach) at this point.
But to keep the response, we use `Response::create` to recreate the response

spire-mike:
@sy-records @driesvints it appears that this timeout doesn't work nicely with atomic locks. For example, with `config.octane.max_execution_time = 30`, and hitting the following route:

```
Route::get('/test-route', function () {
  Cache::lock('test-lock')->get(function () {
    sleep(60);
  });

  return response('');
});
```

The `test-lock` gets set, but it never gets released when the 408 response is sent. Is there some way to 'cleanup' after the request times out?

I'm on Laravel `8.83.11`, Octane `1.2.8` using swoole.

PR #519, opened 2022-05-12
Revert PaginationState to resolve using new instance of the app

farmani:
based on our experience PaginationState will use different instance (old instance) of the app and cannot recognize new parameter has been set in the route it seems this was implemented before but commented out.

this PR revert them back but in Listener class it was using `sandbox` not sure what was the reason I change it to `app`

### DIFF
--- a/src/Listeners/GiveNewRequestInstanceToPaginator.php
+++ b/src/Listeners/GiveNewRequestInstanceToPaginator.php
@@ -14,6 +14,6 @@ class GiveNewRequestInstanceToPaginator
      */
     public function handle($event): void
     {
-        // PaginationState::resolveUsing($event->sandbox);
+        PaginationState::resolveUsing($event->app);
     }
 }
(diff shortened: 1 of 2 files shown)


src/Listeners/GiveNewRequestInstanceToPaginator.php
```diff
@@ -14,6 +14,6 @@ class GiveNewRequestInstanceToPaginator
      */
     public function handle($event): void
     {
-        // PaginationState::resolveUsing($event->sandbox);
+        PaginationState::resolveUsing($event->app);
```
taylorotwell, line 17: Why did you change this to $event->app? Do you understand that change?

farmani: sandbox is a clone of application (I consider it as a copy of the app to switch back) but both of them get a new instance of the request in `GiveNewRequestInstanceToApplication`.

so there should not be any different really which one to use but as still not clear to me what exactly cause this issue I preferred to use `app` beside sandbox because it is most recent copy instance of app with all changes applied on it during the processing of a request.

anyway I revert it back to `sandbox` as it was probably worked before there should not be any issue with it too.

PR #521, opened 2022-05-17
Add Concurrent

huangdijia:
Based on `Swoole\Coroutine\Channel` implementation, it is used to control the maximum number of concurrent coroutines in a code block. In the following example, when 10 coroutines are executed at the same time, they will block in the loop, but only the current coroutine will be blocked until a position is released, and the loop will continue to execute the next coroutine.

```php
use Laravel\Octane\Facades\Octane;

$concurrent = Octane::concurrent(10);

for ($i = 0; $i < 15; ++$i) {
    $concurrent->create(function () {
        // Do something...
    });
}
```

or

```php
use App\Models\User;

$concurrent = Octane::concurrent(100);
User::cursor()->each(function($user) use ($concurrent) {
    $concurrent->create(function() use ($user) {  /* do something*/ });
});
```

### DIFF
--- a/src/Concerns/ProvidesConcurrencySupport.php
+++ b/src/Concerns/ProvidesConcurrencySupport.php
@@ -2,13 +2,13 @@
 
 namespace Laravel\Octane\Concerns;
 
-use Swoole\Http\Server;
+use Laravel\Octane\Contracts\DispatchesTasks;
 use Laravel\Octane\Swoole\Concurrent;
 use Laravel\Octane\Swoole\ServerStateFile;
 use Laravel\Octane\SequentialTaskDispatcher;
-use Laravel\Octane\Contracts\DispatchesTasks;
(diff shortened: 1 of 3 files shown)


foremtehan:
How this compare compare to [Octane::concurrently](https://github.com/laravel/octane/blob/722b7567e2817b3f5e4a4ae7b49290f53968a114/src/Concerns/ProvidesConcurrencySupport.php) ?

huangdijia:
> How this compare to [Octane::concurrently](https://github.com/laravel/octane/blob/722b7567e2817b3f5e4a4ae7b49290f53968a114/src/Concerns/ProvidesConcurrencySupport.php) ?
> 
> 
> 
> Also see https://github.com/laravel/octane/pull/515

Reference examples, higher performance.

taylorotwell:

PR #533, opened 2022-05-27
[1.x] Remove roadrunner rpc default setting

ChanJuiHuang:
I think rpc default open will be removed. There are reasons below,
1. Default open rpc listen will occupy a port number. It wastes resource in the server.
2. Always, we use http server, not tcp server. If someone wants to use tcp server, he or she can set in **.rr.yaml** to enable tcp server.


driesvints:
Isn't this a breaking change?

ChanJuiHuang:
It is breaking change. Do I change the merged branch？
I don't see the version which greater than 1

driesvints:
You need to target master.

PR #534, opened 2022-05-30
Remove roadrunner rpc default setting

ChanJuiHuang:
I think rpc default open will be removed. There are reasons below,
1. Default open rpc listen will occupy a port number. It wastes resource in the server.
2. Always, we use http server, not tcp server. If someone wants to use tcp server, he or she can set in **.rr.yaml** to enable tcp server.


taylorotwell:
I have no idea if this will break existing applications and don't have time to dig into it at the moment.

PR #537, opened 2022-06-07
Add relay dsn option

Kirill-Znamenskiy:


driesvints:
Please re-send this with a thorough explanation why this is needed. Thanks

Kirill-Znamenskiy:
> Please re-send this with a thorough explanation why this is needed. Thanks

Socket or TCP port relay usage is strongly recommended for avoiding problems with dd(), dump(), echo() and other similar functions, that sends data to the IO pipes. 
This explanation is enough or need more?

qindj:
@Kirill-Znamenskiy Could you please re-send this PR? I need this too.

Issue #432, opened 2021-12-05
octane:reload command doesn't reflect changes (swoole)


michael-rubel:
- Octane Version: v1.0.20
- Laravel Version: v8.73.2
- PHP Version: v8.1.0
- Server & Version: Swoole v4.8.2
- Database Driver & Version: MariaDB v10.6.4

### Description:
`octane:reload` doesn't reflect changes with Swoole.
The issue basically the same as [this one](https://github.com/laravel/octane/issues/364) or [this one](https://stackoverflow.com/questions/68299225/octanereload-on-centos-8-doesnt-reflect-changes).
`octane:stop` works correctly, but creates downtime.

### Steps To Reproduce:
- Set up Octane on CentOS 7 with Nginx and Supervisor:
```
[program:octane]
command=/usr/bin/php -d variables_order=EGPCS /var/www/project/artisan octane:start --server=swoole --host=0.0.0.0 --port=8000 --max-requests=10000
autostart=true
autorestart=true
stopasgroup=true
killasgroup=true
user=youruser
stdout_logfile=/var/www/project/storage/logs/octane-workers.log
stopwaitsecs=60
```
Change something in the project, then run the `octane:reload` command from `youruser`.

kevincobain2000:
Yes, I have also noted the same issue.
It doesn't reflect changes when new route/controller is added.

driesvints:
Since this is about a specific bistro on Linux that we don't use ourselves I'm afraid we can't provide support here. It works well on Ubuntu (the default distro on Forge). We'd very much welcome PR's that would run Octane on CentOS.

michael-rubel:
@driesvints Looks like the issue isn't CentOS-related.
It's something about the configuration cache and Octane's internal behavior.
When I type the following:
```
php artisan optimize:clear
php artisan octane:reload
```
Then it works.
If I add:
```
php artisan optimize:clear
php artisan optimize
php artisan octane:reload
```

Then the changes aren't made.

Perhaps Laravel takes the snapshot of the optimized config cache from an existing Octane instance so we cannot update the cache. Maybe it doesn't make sense at all in the case of using Octane?

PR #547, opened 2022-06-25
Moved opcache clearing to before bootWorker

### LINKED ISSUE #546: Opcache is not cleared before worker (re)boot
- Octane Version: v1.2.12
- Laravel Version: v9.3.1
- PHP Version: 8.1.0
- Server & Version: OpenSwoole 4.11 and also Swoole 4.8.10
- Database Driver & Version: Not relevant

### Description:
The opcache is not cleared before the worker boot, resulting in hot reload not working with all bootstrapped content using at least Swoole and OpenSwoole. (tested with both)

The solution is very simple, and is simply a re-ordering of the boot sequence of a worker, moving the [`clearOpcodeCache()`](https://github.com/laravel/octane/blob/1.x/src/Swoole/Handlers/OnWorkerStart.php#L41) method call before the [`$this->bootWorker()`](https://github.com/laravel/octane/blob/1.x/src/Swoole/Handlers/OnWorkerStart.php#L37) method call fixes the issue on my setup.

### Steps To Reproduce:

1. Enable opcache in php.ini: `opcache.enable_cli=1`
2. With the server running in --watch mode, edit any bootstrapped file, for example a routes file.
3. Observe hot reloading is not having any effect on bootstrapped files, for example that routes do not change untill server restart.

LauJosefsen:
Moved opcached clearing to before bootWorker method to enable hot reloading for Swoole and OpenSwoole when Opcache is enabled.

More information about the issue this solves can be seen in #546 .

Closes #546

### DIFF
--- a/src/Swoole/Handlers/OnWorkerStart.php
+++ b/src/Swoole/Handlers/OnWorkerStart.php
@@ -31,6 +31,8 @@ class OnWorkerStart
      */
     public function __invoke($server, int $workerId)
     {
+        $this->clearOpcodeCache();
+
         $this->workerState->server = $server;
         $this->workerState->workerId = $workerId;
         $this->workerState->workerPid = posix_getpid();
@@ -38,7 +40,6 @@ class OnWorkerStart
 
         $this->dispatchServerTickTaskEverySecond($server);
         $this->streamRequestsToConsole($server);
-        $this->clearOpcodeCache();
 
         if ($this->shouldSetProcessName) {
             $isTaskWorker = $workerId >= $server->setting['worker_num'];


michael-rubel:
Looks like it's related to this issue: https://github.com/laravel/octane/issues/432

LauJosefsen:
> Looks like it's related to this issue: https://github.com/laravel/octane/issues/432

It sure does seem like that. Also the issues linked from that issue. (funny how i didn't find those by searching for issues with "--watch" 🤔) 

I believe the change I propose is both safe and fixes the issue completely.

binaryfire:
Sorry to bump an old thread, but is anyone still having this issue? I'm using Swoole and reloading workers (either automatically via --watch / Chokidar or manually via `php artisan octane:reload` doesn't always reflect changes. The only way to get it working is via `php artisan octane:stop`. 

If anyone has a workaround they could share it would be much appreciated. I'm using Docker but I don't think that should make a difference- everything is run in the container.

LauJosefsen:
@binaryfire Are you using file opcache by chance? Maybe link your php.ini file? We do not have any problems anymore, also using Swoole on our deployments.

binaryfire:
@LauJosefsen Hi Lau, thanks for the reply. Yeah I'm using opcache. Here's the link to the PHP config I'm using in the container: https://gist.github.com/binaryfire/ed1ec34f8333117aa7e494192f807aa2

It's identical to Sail: PHP 8.2 from the Ondrej PPA on Ubuntu 22.04, stock php.ini with the same custom config in `conf.d` folder. Not sure what's going on! I've considered implementing a custom listener or file watcher to fully stop and start Octane on file changes, but I'm hoping there's a more elegant solution.

PR #549, opened 2022-07-04
Fix http statuses

vostlertsd:


GrahamCampbell:
What does this fix please? This breaks our tests.

driesvints:
This seems to break the test suite.

PR #555, opened 2022-07-22
Resolve custom vendor folder

oleksiikhr:
https://getcomposer.org/doc/03-cli.md#composer-vendor-dir

Support of a running worker using a custom vendor folder.

### DIFF
--- a/bin/bootstrap.php
+++ b/bin/bootstrap.php
@@ -44,7 +44,9 @@ if (! function_exists('dd')) {
 |
 */
 
-if (! is_file($autoload_file = $basePath.'/vendor/autoload.php')) {
+$vendorPath = getenv('COMPOSER_VENDOR_DIR') ?: $basePath.'/vendor';
+
+if (! is_file($autoload_file = $vendorPath.'/autoload.php')) {
     fwrite(STDERR, "Composer autoload file was not found. Did you install the project's dependencies?".PHP_EOL);
 
     exit(10);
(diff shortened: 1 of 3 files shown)


taylorotwell:

oleksiikhr:
@taylorotwell Thanks for the reply, sorry for not explaining the problem right away.

I currently have a custom architecture that looks something like this:

```
- apps
  |-- src
  |-- ui
      |-- app1
          |-- bin
          |-- bootstrap
          |-- config
          |-- public
          |-- ...
      |-- app2
      |-- app3
- src (DDD architecture)
- vendor
```

So, I have several projects that work on a single code base.

In each application:

- base_path: path to `apps/ui/{project}`.
- namespace: overriden in `Illuminate\Foundation\Application`.

I.e. everything works fine except laravel/octane, which has a hardcoded path to the vendor folder.

---

Even if I don't run the roadrunner with the artisan command, I still can't do it, because inside `bin/bootstrap` it tries to take `vendor` from `base_path`, which I don't have.

---

Similar Pull Requests:

- https://github.com/laravel/framework/pull/41463
- https://github.com/laravel/tinker/pull/73

oleksiikhr:
@taylorotwell Sorry for another ping, but in case you didn't see the previous message.

If the changes look too extensive, can we make edits only in the `bin/bootstrap.php` file? That would be enough to be able to run `bin/roadrunner-worker` or `bin/swoole-server` with a different path to the `vendor` folder without linking to the laravel base path.

That would already be enough to implement, at least with my configs/commands.

Can we make these changes?

If necessary, I can make a test repository to show the problem.

Please, let me know if we can't do that.

gawsoftpl:
I agree with @oleksiikhr. Octane dont work correct with monorepo without custom vendor dir path. This path in main repo will be great. Without this I have make manualy link to vendor directory in staging and in production

PR #558, opened 2022-08-02
Fixed starting RoadRunner on a readonly filesystem

vokamut:
Fixed:
`PHP Warning:  touch(): Unable to create file /var/www/.rr.yml because Permission denied`

### DIFF
--- a/src/Commands/StartRoadRunnerCommand.php
+++ b/src/Commands/StartRoadRunnerCommand.php
@@ -145,7 +145,7 @@ class StartRoadRunnerCommand extends Command implements SignalableCommandInterfa
 
         if (! $path) {
             $path = base_path('.rr.yaml');
-            
+
             touch($path);
         }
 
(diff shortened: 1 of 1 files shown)


zcsimon:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

src/Commands/StartRoadRunnerCommand.php
```diff
@@ -144,9 +144,9 @@ protected function configPath()
         $path = $this->option('rr-config');
 
         if (! $path) {
```
Namoshek, line 146: By introducing a default value for the `--rr-config` option, this part is never called anymore. The default `.rr.yaml` is now also not wrapped with `base_path()` anymore, so the check on L152 should actually fail for fresh installs.

Since your problem seems to be container related, why not simply add `RUN touch /var/www/.rr.yaml` in your `Dockerfile` to match the fallback behavior on L147 without granting too many permissions?

vokamut: But if the file is not specified, then when running the `php artisan octane:start` code updates the access time via `touch()`, and the file is not allowed to be changed

taylorotwell:
I am closing this pull request because it lacks sufficient explanation, tests, or both. It is difficult for us to merge pull requests without these things because the change may introduce breaking changes to the framework.

Feel free to re-submit your change with a **thorough** explanation of the feature and tests - integration tests are preferred over unit tests. Please include it's benefit to end users; the reasons it does not break any existing features; how it makes building web applications easier, etc.

Thanks!

PR #563, opened 2022-08-08
This PR fixes #562 by defining "non-standard HTTP status codes reasons" in the octane config file.

### LINKED ISSUE #562: Changing non-standard HTTP status codes to 200 OK
- Octane Version: 1.3.0
- Laravel Version: 9.19
- PHP Version: 8.1.9
- Server & Version: Swoole 4.11.1

### Description:
This is so similar to issue #291, just instead of `Illuminate\Session\TokenMismatchException` we used `Symfony\Component\HttpKernel\Exception\HttpException`

When I try to use any non-standard HTTP status codes I always get "200 OK", but in Octane logs its right status code, for example: 

```
Local: http://0.0.0.0:8801 

599    GET /health/ping ................................... 28.99 mb 43.05 ms
```

### Steps To Reproduce:
1. Start the octane 
2. Call the route from Postman
3. Check that the responded status code is 200 instead of 5xx

saderi:
This PR fixes #562 by defining "non-standard HTTP status codes reasons" in the octane config file.

This PR allows users to define other status code reasons in `octane.php` config file

### DIFF
--- a/config/octane.php
+++ b/config/octane.php
@@ -231,8 +231,6 @@ return [
 
     'status_code_reasons' => [
         419 => 'Page Expired',
-        431 => 'Request Header Fields Too Large',                // RFC6585
-        451 => 'Unavailable For Legal Reasons',                  // RFC7725
     ],
 
 ];
--- a/config/octane.php
+++ b/config/octane.php
@@ -224,7 +224,7 @@ return [
     |--------------------------------------------------------------------------
     | The following setting allows you to set non-standard HTTP status codes
     | for Octane. Octane gonna return "200 OK" in responded status if you
-    | use a "non-standard HTTP status codes" in your response header. 
+    | use a "non-standard HTTP status codes" in your response header.
     |
     */
 
--- a/config/octane.php
+++ b/config/octane.php
@@ -218,18 +218,4 @@ return [
 
     'max_execution_time' => 30,
 
-    /*
-    |--------------------------------------------------------------------------
-    | Status Code Reasons for Octane Swoole
-    |--------------------------------------------------------------------------
-    | The following setting allows you to set non-standard HTTP status codes
-    | for Octane. Octane gonna return "200 OK" in responded status if you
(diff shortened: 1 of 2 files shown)


src/Swoole/SwooleClient.php
```diff
@@ -19,12 +19,6 @@
 
 class SwooleClient implements Client, ServesStaticFiles
 {
-    const STATUS_CODE_REASONS = [
-        419 => 'Page Expired',
-        431 => 'Request Header Fields Too Large',                             // RFC6585
-        451 => 'Unavailable For Legal Reasons',                               // RFC7725
-    ];
-
```
driesvints, line 27: Removing this here is a breaking change.

saderi: @driesvints I made a small change to fix this

config/octane.php
```diff
@@ -218,4 +218,19 @@
 
     'max_execution_time' => 30,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Status Code Reasons for Octane Swoole
+    |--------------------------------------------------------------------------
+    |
+    | If you use non-standard HTTP status codes in your response, Octane gonna
```
nunomaduro, line 226: Can you apply the style of "3 characters less than the previous line" in this comment please?

nunomaduro (changes requested):
Can you add tests to this pull request please?

saderi:
> Can you add tests to this pull request please?

I can update current test if its OK, or write a new one.

saderi:
@nunomaduro Is [this test](https://github.com/laravel/octane/blob/ef741f3f73a784e5e75293f46bb34daba878deac/tests/SwooleClientTest.php#L193) not acceptable for this PR?

taylorotwell:
I don't really love calling the global `config` helper from here. IMO just return a standard status and include a JSON payload providing a more detailed error code / status.

LordVersA:
@taylorotwell can you provide any other option to have other custom status codes (like 481) instead of using configuration file?

PR #564, opened 2022-08-11
Output Swoole request information based on swoole log_level

L3o-pold:
Before this PR Swoole request logs are only display if app is in local or testing environment. 

With this PR it display request logs based on the Swoole log_level. There is no change for the user with default configuration.


taylorotwell:
I would need more community feedback and review on this to be honest.

benavern:
The fact that we can't have logs on production is something that prevents us to be able to find easily some errors that occure only on production :( 

I think that it should be something that we should be able to activate by ourselves (using log_level or any other environment variable)

ignasbernotas:
I have been waiting for this for quite some time, having request logs can help us match events between multiple applications. In any case, should be up to the user to decide how to configure this.

PR #569, opened 2022-09-01
Update link to OpenSwoole website

spawnia:
- https://swoole.co.uk redirects to https://openswoole.com
- https://swoole.com is linked at https://github.com/swoole/swoole-src

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


driesvints:
Most of the swoole.com site is in Chinese so I don't think it makes sense to link to it by default.

driesvints:
Thanks!

PR #574, opened 2022-09-09
make STATUS_CODE_REASONS public static

raftx24:
Based on the [swoole documentation](https://openswoole.com/docs/modules/swoole-http-response-status#description), swoole will ignore `custom status code` without `reason`. 

> If an invalid status code is set, Swoole will reset the value to 200. However, if a HTTP status code $reason is given, you may set any status code such as 499 or 856.

I changed STATUS_CODE_REASON to public static so other projects also register their custom status code.

### DIFF
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -19,7 +19,7 @@ use Throwable;
 
 class SwooleClient implements Client, ServesStaticFiles
 {
-    const STATUS_CODE_REASONS = [
+    public static $STATUS_CODE_REASONS = [
         419 => 'Page Expired',
         431 => 'Request Header Fields Too Large',                             // RFC6585
         451 => 'Unavailable For Legal Reasons',                               // RFC7725


driesvints:
This is a breaking change.

raftx24:
@driesvints how? It only adds an option for changing `STATUS_CODE_REASONS` from outside

raftx24:
@driesvints I fixed the tests.

driesvints:
You're changing the constant to a static variable.

raftx24:
@driesvints Ok, I understand what you are saying. Just consider that in octane it wasn't used anywhere else, and probably no one used it anywhere else in other project.
but do you agree with this approach?
1. adding new property.
```
    private static $CUSTOM_STATUS_CODE = [];
    
    public static function addCustomStatusCode(int $code, string $reason)
    {
        static::$CUSTOM_STATUS_CODE[$code] = $reason;
    }
```

and changing `getReasonFromStatusCode` to this:

```diff

    /**
     * Get the HTTP reason clause for non-standard status codes.
     *
     * @param  int  $code
     * @return string|null
     */
    protected function getReasonFromStatusCode(int $code): ?string
    {
-        if (array_key_exists($code, self::STATUS_CODE_REASONS)) {
+        if (array_key_exists($code, array_merge(self::STATUS_CODE_REASONS, static::$CUSTOM_STATUS_CODE))) {
            return self::$STATUS_CODE_REASONS[$code];
        }

        return null;
    }
```

Unfortunately, there is no way to customize this constant in octane, and we really need to change it.
thank you.

driesvints:
Feel free to attempt a new PR 👍

PR #576, opened 2022-09-09
Handle worker startup errors for development purposes

MadridianFox:
## Context
When swoole server running with --watch option, saving of broken code causes shutdown of master process.
It is may be reasonable for production environment, but in development I want to see the changes right after error is fixed, without manual restart of the server process.

## What did i do
Add option `--dont-shutdown-on-error` for the `octane:swoole` command. With this option `OnWorkerStart` event listener doesn't stops the server, but returns an instance of special dummy worker, which aimed for exception rendering.


taylorotwell:

PR #577, opened 2022-09-09
[1.x] Gracefully stop Swoole when running octane:stop

### LINKED ISSUE #469: Terminate gracefully on SIGTERM
- Octane Version: 1.2.0
- Laravel Version: 8.80.0
- PHP Version: 8.0.9
- Server & Version: Swoole 4.7.0
- Database Driver & Version: None

### Description:

When terminating a running Octane process with `SIGTERM`, it exits with the non-zero exit code `137`.
This indicates a crash or abrupt failure.

I would expect the process to terminate gracefully with exit code `0`.

### Steps To Reproduce:

```sh
root@59d2aec7e8b2:/var/www# php artisan octane:start 

   INFO  Server running…

  Local: http://127.0.0.1:8000 

root@59d2aec7e8b2:/var/www# echo $?
137
```

In a second terminal, while Octane is running:

```sh
root@59d2aec7e8b2:/var/www# ps aux | grep octane
root       264  0.1  0.2 119612 59224 pts/1    S+   15:34   0:00 php artisan octane:start
root       307  0.0  0.0   3084   820 pts/2    S+   15:36   0:00 grep octane
root@59d2aec7e8b2:/var/www# kill -TERM 264
```

LukeAbell:
Fixes #469

The existing behavior when running `octane:stop` is to send a `SIGKILL` signal to the Swoole master process, manager process, and all of the worker processes. This causes in-flight requests to immediately be terminated instead of waiting for them to complete.

This PR changes the behavior of `ServerProcessInspector::stopServer()` and `SignalDispatcher::terminate()` to support gracefully shutting down the workers and master/manager process so that all existing requests can complete (with a timeout) before terminating.

The revised flow in this PR:
- `octane:stop` is called (either via artisan or when a SIGTERM/SIGINT is signaled)
- Send `SIGTERM` to all workers
- Wait for all workers to terminate (With a max of the `octane.max_execution_time` config value)
- Send `SIGTERM` to the master and manager process.

Although not used in the octane package as far as I can tell, backwards compatibility of the `SignalDispatcher::terminate()` is maintained.

### DIFF
--- a/src/OctaneServiceProvider.php
+++ b/src/OctaneServiceProvider.php
@@ -61,6 +61,7 @@ class OctaneServiceProvider extends ServiceProvider
                 $app->make(SignalDispatcher::class),
                 $app->make(SwooleServerStateFile::class),
                 $app->make(Exec::class),
+                $app['config']->get('octane.max_execution_time', 30)
             );
         });
 
--- a/src/Swoole/ServerProcessInspector.php
+++ b/src/Swoole/ServerProcessInspector.php
@@ -10,6 +10,7 @@ class ServerProcessInspector
         protected SignalDispatcher $dispatcher,
         protected ServerStateFile $serverStateFile,
         protected Exec $exec,
+        protected int $terminateWait,
     ) {
     }
 
@@ -58,8 +59,12 @@ class ServerProcessInspector
 
         $workerProcessIds = $this->exec->run('pgrep -P '.$managerProcessId);
 
-        foreach ([$masterProcessId, $managerProcessId, ...$workerProcessIds] as $processId) {
-            $this->dispatcher->signal((int) $processId, SIGKILL);
+        foreach ($workerProcessIds as $processId) {
+            $this->dispatcher->signal((int)$processId, SIGTERM);
+        }
+
+        foreach ([$managerProcessId, $masterProcessId] as $processId) {
+            $this->dispatcher->terminate((int)$processId, $this->terminateWait);
         }
 
         return true;
--- a/src/Swoole/ServerProcessInspector.php
+++ b/src/Swoole/ServerProcessInspector.php
@@ -59,13 +59,8 @@ class ServerProcessInspector
 
         $workerProcessIds = $this->exec->run('pgrep -P '.$managerProcessId);
 
-        foreach ($workerProcessIds as $processId) {
-            $this->dispatcher->signal((int)$processId, SIGTERM);
-        }
-
-        foreach ([$managerProcessId, $masterProcessId] as $processId) {
-            $this->dispatcher->terminate((int)$processId, $this->terminateWait);
-        }
+        $this->dispatcher->terminate($workerProcessIds, $this->terminateWait - 5);
+        $this->dispatcher->terminate([$managerProcessId, $masterProcessId], 5);
 
         return true;
     }
--- a/src/Swoole/ServerProcessInspector.php
+++ b/src/Swoole/ServerProcessInspector.php
@@ -59,8 +59,8 @@ class ServerProcessInspector
 
         $workerProcessIds = $this->exec->run('pgrep -P '.$managerProcessId);
 
-        $this->dispatcher->terminate($workerProcessIds, $this->terminateWait - 5);
-        $this->dispatcher->terminate([$managerProcessId, $masterProcessId], 5);
+        $this->dispatcher->terminate($workerProcessIds, $this->terminateWait);
+        $this->dispatcher->terminate([$managerProcessId, $masterProcessId]);
 
         return true;
     }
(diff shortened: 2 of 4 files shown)


LukeAbell:
Drafting - while testing in production we've discovered this works correctly with OpenSwoole but not Swoole. Looking for a solution that works with both Swoole/OpenSwoole

driesvints:
Feel free to resend this when you have more time.

LukeAbell:
@driesvints Will do. We're still working to figure out how to get Octane and our Google Kubernetes setup to gracefully scale. I'll re-submit when we've got things nailed down.

fannyfan414:
> @driesvints Will do. We're still working to figure out how to get Octane and our Google Kubernetes setup to gracefully scale. I'll re-submit when we've got things nailed down.

Do you have any solutions for swoole?

phillip-elm:
@driesvints:

Just making sure I understand this correctly - is the current accepted behavior in Octane to just kill in-progress requests without any grace period?

I see that you closed both issues pertaining to this topic (this and #469). I'm coming from a Go background, but this would seem to be a basic and important issue to address - so I'm a bit confused as to why these were closed without further planning. 

Are there further plans to address it?

Also, @LukeAbell - did anything come of your setup? Any recommendations for handling this issue while it's getting sorted?

driesvints:
Right now we don't have any plans to address this ourselves but we're still open to propositions through PRs. I'm not sure myself what the best way to handle this would be so please attempt a PR and let's go from there 👍

phillip-elm:
For anyone who's on Kubernetes or similar orchestrators and is interested in a workaround:

**Context:**

- Kubernetes shuts down a pod by rerouting traffic away from the container, calling the `PreStop` hook if present, then sends `SIGTERM`.
- Normally processes would take the time they need to finish requests / tasks after receiving `SIGTERM`, but Octane just dies immediately.

**Workaround:**

Add a reasonable delay to the `PreStop` hook. This assumes you're keeping API / web responses to <1 second and moving longer tasks to workers.

Requests taking longer than usual that exceed the delay period would still be killed prematurely by this workaround.

**Alternative:**

If your setup doesn't allow this, an alternative would be to write a forking wrapper that intercepts `SIGTERM`, waits for the desired delay period, then sends `SIGTERM` to the child before terminating.

smortexa:
@LukeAbell Do you have any plan to resend this PR?

LukeAbell:
@smortexa @phillip-elm @fannyfan414 I don't currently plan to submit a revised PR. We've been using my [fork](https://github.com/LukeAbell/octane/tree/gracefully-stop-swoole) for 6 months on a high traffic complex Laravel application Kubernetes environment using OpenSwoole and it's completely resolved our issues when pods are terminated. 

We have some unique requirements that require our timeout to be more than a few seconds so just adding a sleep via the `PreStop` hook wasn't a viable workaround.

I don't know enough about the internals of Swoole vs Open Swoole to figure out why they behave differently with my changes, but the fork works great with Open Swoole. I don't think there's any downside to merging this into Octane core even though there's no improvement for Swoole (only Open Swoole). Not sure if that's an option though @driesvints?

driesvints:
@sy-records do you have any thoughts here?

sy-records:
Does `max_wait_time` work?

```php
    'swoole' => [
        'options' => [
            'max_wait_time' => 5,
        ],
    ],
```

smortexa:
@sy-records No, It doesn't.

LauJosefsen:
If OpenSwole and Swoole are moving in different directions, maybe it's worth separating the Octane implementation into two, such as it currently is with RoadRunner and Swoole already? Then have three options, RoadRunner, Swoole, and OpenSwoole, which would then allow the graceful shutdown of OpenSwoole of this PR to be merged?

taylorotwell:
Does someone have time to explore supporting OpenSwoole?

taylorotwell:
Or, can we not determine if OpenSwoole is being used and use the graceful restart approach in that scenario?

LukeAbell:
I can implement something to gracefully shutdown if OpenSwoole is detected. Anyone know how to detect Swoole vs OpenSwoole?

smortexa:
@LukeAbell https://github.com/laravel/octane/blob/3a9f2382a76ec5ee66deda1d3974a3f4d741e488/src/Tables/TableFactory.php#L16

xorinzor:
Can't this just be merged with a note in the docs stating that only OpenSwoole currently supports the graceful shutdown?

Seems like a shame to leave this out only because Swoole doesn't support this yet. I'd argue it's better to leave the decision with the developer to optionally switch to OpenSwoole if they require this feature. At least that way they'd have the option to do so instead of defaulting with all kinds of workarounds.

arsengoian:
Any status update in 2024 whether this started working on Swoole yet?

PR #579, opened 2022-09-12
prevent ob_end_clean(): Failed to delete buffer. No buffer to delete error 

zack7z:
i added a condition to prevent  "ob_end_clean(): Failed to delete buffer. No buffer to delete" error in some situation (in my case ckfinder)


zack7z:
g

PR #589, opened 2022-09-29
[1.x] Flushes controller on route

nunomaduro:


olivernybroe:
I have tested out if this fix together with dev laravel branch resolves the issue and it does indeed. 🎉

Issue #592, opened 2022-10-03
Serving static files with non-ASCII characters naming


smortexa:
- Octane Version: 1.3.1
- Laravel Version: 9.29.0
- PHP Version: 8.1.8
- Server & Version: Swoole 4.8.10
- Database Driver & Version: N/A

### Description:
The Laravel Octane Swoole HTTP server can't serve static files with non-ASCII characters naming and returns Laravel 404 page.

### Steps To Reproduce:
 
Put `تست.jpg` or `テスト.jpg` in `public` directory and hit 
`<your-local-server>/تست.jpg` or `<your-local-server>/テスト.jpg`.
You will face with Laravel 404 error page.

NathanFreeman:
Can you please show me you the setting of `document_root` , `enable_static_handler` ?

smortexa:
@NathanFreeman I didn't add any extra configuration to Octane's default configuration, and Octane has no specific config for `document_root` and `enable_static_handler`.

sy-records:
This is a configuration item provided by Swoole, which you can set in `config/octane.php`

```php
    'swoole' => [
        'options' => [
            'document_root' => public_path(),
            'enable_static_handler' => true,
        ],
    ],
```

smortexa:
@sy-records I can confirm that with this configuration the issue was resolved.

driesvints:
Glad you found a solution. Thanks @sy-records

PR #593, opened 2022-10-04
Handle static files by Swoole HTTP server

smortexa:
This PR follows #592 and configures static files to be handled by the Swoole HTTP server, not Laravel.

This will:
- Fix serving static files with non-ASCII characters naming issues
- Increase the app's throughput
- Reduce resource consumption


taylorotwell:
Didn't this comment show how to opt-in to this already?

https://github.com/laravel/octane/issues/592#issuecomment-1266317765

smortexa:
NO. IMO, supporting non-ASCII characters shouldn't be opt-in.

PR #605, opened 2022-10-31
[1.x] Read port for http server from environment if no port is passed 

hendrikheil:
This pull request aims to add support for defining the HTTP server port through environment variables. The implementation shouldn't break any existing behavior.
This feature should be helpful for anyone that needs to run octane with a randomly generated port.

### DIFF
--- a/src/Commands/StartCommand.php
+++ b/src/Commands/StartCommand.php
@@ -16,7 +16,7 @@ class StartCommand extends Command implements SignalableCommandInterface
     public $signature = 'octane:start
                     {--server= : The server that should be used to serve the application}
                     {--host=127.0.0.1 : The IP address the server should bind to}
-                    {--port=8000 : The port the server should be available on}
+                    {--port= : The port the server should be available on}
                     {--rpc-port= : The RPC port the server should be available on}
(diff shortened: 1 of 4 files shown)


src/Commands/StartCommand.php
```diff
@@ -16,7 +16,7 @@ class StartCommand extends Command implements SignalableCommandInterface
     public $signature = 'octane:start
                     {--server= : The server that should be used to serve the application}
                     {--host=127.0.0.1 : The IP address the server should bind to}
-                    {--port=8000 : The port the server should be available on}
```
Namoshek, line 19: @hendrikheil  This change is safe but slightly inconvenient as it removes the default port from the `--help` text of the command:
```
// Before
--port[=PORT]                  The port the server should be available on [default: "8000"]

// After
--port[=PORT]                  The port the server should be available on
```

Might be worth to add this to the text :+1:

driesvints: @Namoshek good suggestion. Added that: https://github.com/laravel/octane/blob/1.x/src/Commands/StartCommand.php

PR #607, opened 2022-11-02
[1.x] Make IP to bind the server to configurable via env

xorinzor:
Building further on #605 this also makes the IP that the server listens on configurable using environment variables when no commandline argument for `--host` is provided.

The implementation should not break any existing deployments.


xorinzor:
Ideally all of these arguments should be configurable via the `octane` config file. However I discovered quite quickly that this would break backwards compatibility with existing deployments unless they updated their config files.

Perhaps this is a breaking change worth considering for 2.0 and add it to a migration guide.

Issue #610, opened 2022-11-08
rr binary has wrong permissions after downloading


Dylan-Buth:
- Octane Version: 1.3.5
- Laravel Version: 9.37.0
- PHP Version: 8.1
- Server & Version: RoadRunner 2.11.4
- Database Driver & Version:

### Description:
Just upgraded to Laravel 9 and PHP 8.1, for Laravel 8 and PHP 8.0. We use docker for local development as well as production. When starting the php container after the upgrade it exits immediately. Output in screenshot below.

<img width="1097" alt="image" src="https://user-images.githubusercontent.com/2185074/200640091-53c6b25c-8479-4822-b084-b09562b034a0.png">

The rr binary that gets downloaded has permissions set to `--wxrw--wt`. I know the `downloadRoadRunnerBinary()` method is being ran, but it seems as if the `chmod()` step isn't running successfully. If I manually set the permissions to 755 it works just fine. 

<details>
<summary>Dockerfile for php container is below</summary>

``` 
FROM composer:2.1 as composer

FROM php:8.1-cli-alpine
COPY --from=composer /usr/bin/composer /usr/bin/composer
COPY ./deploy/php/docker-php-entrypoint /usr/local/bin/

RUN apk add --update --no-cache autoconf automake
RUN apk add --update --no-cache build-base
RUN apk add --update --no-cache mysql-client
RUN apk add --update --no-cache openssl
RUN apk add --update --no-cache gmp-dev
RUN apk add --update --no-cache zlib-dev
RUN apk add --update --no-cache libzip-dev
RUN docker-php-ext-install gmp
RUN docker-php-ext-install pdo_mysql
RUN docker-php-ext-install zip
RUN CFLAGS="$CFLAGS -D_GNU_SOURCE" docker-php-ext-install sockets
RUN docker-php-ext-install pcntl

RUN pecl install -o -f redis \
    &&  rm -rf /tmp/pear \
    &&  docker-php-ext-enable redis

WORKDIR /var/www

# OPcache defaults
ENV PHP_OPCACHE_ENABLE="1"
ENV PHP_OPCACHE_MEMORY_CONSUMPTION="128"
ENV PHP_OPCACHE_MAX_ACCELERATED_FILES="10000"
ENV PHP_OPCACHE_REVALIDATE_FREQUENCY="0"
ENV PHP_OPCACHE_VALIDATE_TIMESTAMPS="0"

RUN docker-php-ext-install opcache
[...]

PR #611, opened 2022-11-09
Prefix rr binary permissions with 0 to specify it as an octal value

Dylan-Buth:
This resolves the issue described [here](https://github.com/laravel/octane/issues/610).

On php 8.1 octane errors out because it's unable to start the rr binary that gets downloaded. It's unable to because it doesn't have permission, because the `chmod()` that gets run after the binary is installed, doesn't have the right value for permissions. It's trying to set them to `755` but per the [chmod documentation](https://www.php.net/manual/en/function.chmod.php), you need to use a `0` prefix to specify it as an octal value. 

This PR adds the `0` prefix to the permissions, setting them to `0755` which runs successfully on php 8.1.

### DIFF
--- a/src/Commands/Concerns/InstallsRoadRunnerDependencies.php
+++ b/src/Commands/Concerns/InstallsRoadRunnerDependencies.php
@@ -171,7 +171,7 @@ trait InstallsRoadRunnerDependencies
             fn ($type, $buffer) => $this->output->write($buffer)
         );
 
-        chmod(base_path('rr'), 755);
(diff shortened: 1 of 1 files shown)


driesvints:
@Dylan-Buth please see the pr template and don't just link to an issue. Provide a thorough explanation why this is needed for the person who will review this.

Dylan-Buth:
@driesvints Updated the description

kluevandrew:
Is it released?

driesvints:
Release day is on Tuesday.

PR #614, opened 2022-11-16
fix: resolve typo resulting in deprecation notice

owenvoke:
We noticed that on PHP 8.1, we are receiving the following deprecation notice in Sentry.
```txt
json_encode(): Passing null to parameter #2 ($flags) of type int is deprecated in /var/www/vendor/laravel/octane/src/Commands/Concerns/InteractsWithIO.php on line 237
```

It looks like this is just a typo, and the verbosity should be passed to `info()` not to `json_encode()`. 👍🏻

### DIFF
--- a/src/Commands/Concerns/InteractsWithIO.php
+++ b/src/Commands/Concerns/InteractsWithIO.php
@@ -234,7 +234,7 @@ trait InteractsWithIO
             'request' => $this->requestInfo($stream, $verbosity),
             'throwable' => $this->throwableInfo($stream, $verbosity),
             'shutdown' => $this->shutdownInfo($stream, $verbosity),
-            default => $this->info(json_encode($stream, $verbosity))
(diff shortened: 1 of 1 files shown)


driesvints:
@owenvoke this PR seems to have broken our builds. Can you send in a fix for that?

owenvoke:
@driesvints, are you meaning the tests? They were broken for 4 days before my PR. 🤔 I'm not entirely sure what the issue is, will see if I can take a look. 👍🏻

driesvints:
Weird as they seem to start breaking with your PR:

image

owenvoke:
Yeah, the scheduled daily tests were failing. 👀
https://github.com/laravel/octane/actions/workflows/tests.yml

Will look into it shortly.

driesvints:
@owenvoke I see. Thanks for pointing that out and looking into it!

PR #616, opened 2022-11-17
fix: ensure Testbench `vendor` symlink exists

owenvoke:
This should fix the issues with the Testbench `vendor` directory not existing. When any `testbench` command is run, it creates the symlink if it does not exist.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,6 +43,11 @@
             "Laravel\\Octane\\Tests\\": "tests"
         }
     },
+    "scripts": {
+        "post-autoload-dump": [
+            "@php vendor/bin/testbench package:discover --ansi"
+        ]
+    },
     "extra": {
         "branch-alias": {
             "dev-master": "1.x-dev"


driesvints:
Hmm, I'm not sure why this is needed. Was there a breaking change in Testbench? /cc @crynobone

crynobone:
Can I see the original bug report?

owenvoke:
I can't actually see any change in Testbench. But in the last 4 days tests here have been failing ([example 1](https://github.com/laravel/octane/actions/runs/3453132650), [example 2](https://github.com/laravel/octane/actions/runs/3484262475)) because the `vendor/orchestra/testbench-core/laravel/vendor/autoload.php` file does not exist.

This comes from: https://github.com/laravel/octane/blob/64e5884436df84e7211cd4f0ace729ce92aa2768/bin/bootstrap.php#L48

This results in the output:
> Composer autoload file was not found. Did you install the project's dependencies?

driesvints:
I really doubt we need to do this change and that there's another culprit happening. None of our other builds that use Testbench are failing.

crynobone:
`vendor` was never available by default until you run `vendor/bin/testbench`. 

I'm not sure how previously it was "available" without  running the command. So, it maybe some other tests triggering it etc.

crynobone:
Okay, found the root cause.

In tests environments requiring `vendor/orchestra/testbench/laravel/boostrap/app.php` triggers creating `vendor` command in Testbench 7.11 and below. It completed via https://github.com/laravel/octane/blob/722b7567e2817b3f5e4a4ae7b49290f53968a114/src/ApplicationFactory.php#L24-L33

and triggered https://github.com/laravel/octane/blob/64d15b2477c5bd98b05c8e040a8a6831f922e25d/tests/ApplicationFactoryTest.php#L18-L21

In Testbench 7.12, we remove this because depending on `getcwd()` may return unexpected behavior depending on context. https://github.com/orchestral/testbench-core/compare/v7.11.0...v7.12.0#diff-96e6ce0de4b0bfc3bcf8d50f6d0c1089d9348960fe941b5ec56944860bfe316c

------

However, I do think we should use the recommended installation guide to cover enabling `vendor` folder: https://packages.tools/testbench/getting-started/introduction.html#installation

owenvoke:
@crynobone, thank you for looking into this. I've pushed a change that adds the `post-autoload-dump` hook, and this appears to have worked locally. 👍🏻 Will see if it works in CI too. Thank you.

crynobone:
👍 Looks good now

composer.json
```diff
@@ -43,6 +43,11 @@
             "Laravel\\Octane\\Tests\\": "tests"
         }
     },
+    "scripts": {
+        "post-autoload-dump": [
+            "@php vendor/bin/testbench package:discover --ansi"
```
driesvints, line 48: This won't work and will fail on production as testbench isn't available. I think we need to do some things from the docs below? 

https://packages.tools/testbench/basic/testcase.html#package-auto-discovery

owenvoke: This script won't be used on production though, will it? 🤔 And in production, people would be using a path to their app, so it wouldn't be an issue? Or am I missing something?

driesvints: Right, my bad!

PR #630, opened 2023-01-09
Require "laravel/framework" using `--dev` for build

crynobone:
The project `composer.json` already contain `laravel/framework` under `require` and having this line replace the "locked" supported versions.

These changes would also allow us to add `--prefer-lowest` build.


driesvints:
I guess this can't hurt 👍

PR #631, opened 2023-01-12
[1.x] Reset prevent queries state

timacdonald:
Dependent on https://github.com/laravel/framework/pull/45603 being accepted, merged, and tagged.

This PR will always reset the "preventQueries" state to `false`. This does mean, unfortunately, that setting `DB::preventQueries()` in a service provider is not supported. Considering this would probably make your app unusable anyway, I don't think we need to be concerned with this.

In the rare case that functionality was wanted, it would be possible via event listeners instead.

I've deviated from the naming convention as I didn't feel it was "flushing" anything.


nunomaduro:
@timacdonald I am closing this pull request due to a lack of activity. However, please don't hesitate to reopen it once you are ready to proceed.

PR #642, opened 2023-02-03
Fix unsetting a cookie value when using Swoole

### LINKED ISSUE #641: Swoole deprecation errors for cookies with null values
- Octane Version: 1.3.1
- Laravel Version: 9.11
- PHP Version: 8.1.9
- Server & Version: Swoole 5.0.1

### Description:
Out of the box, when you do `cookie()->forget('cookie_name')` Laravel saves a cookie with `null` value under the hood. This cookie is then fed into Swoole's response object here: https://github.com/laravel/octane/blob/1.x/src/Swoole/SwooleClient.php#L192, which in turn triggers a deprecation error from Swoole itself: `Swoole\Http\Response::cookie(): Passing null to parameter #2 ($value) of type string is deprecated in /app/vendor/laravel/octane/src/Swoole/SwooleClient.php:200`.

When either not using Octane at all or using Roadrunner, [Symfony's `Cookie` class](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/HttpFoundation/Cookie.php#L254) handles missing value by converting `null` cookie value to "deleted".

### Steps To Reproduce:

Enable [deprecation warnings](https://laravel.com/docs/9.x/logging#logging-deprecation-warnings); then do `cookie()->forget('cookie_name')` in an Octane Swoole server.

oprypkhantc:
Closes #641 

Following the Symfony's implementation, any empty string or `null` value will instead use string `deleted`. To avoid issues with strings like `0` (which are valid cookie values) I'm not using the `?:` op.

### DIFF
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -193,7 +193,7 @@ class SwooleClient implements Client, ServesStaticFiles
 
             $swooleResponse->{$cookie->isRaw() ? 'rawcookie' : 'cookie'}(
                 $cookie->getName(),
-                !$shouldDelete ? $cookie->getValue() : 'deleted',
+                $shouldDelete ? 'deleted' : $cookie->getValue(),
                 $cookie->getExpiresTime(),
                 $cookie->getPath(),
                 $cookie->getDomain() ?? '',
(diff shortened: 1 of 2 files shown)


taylorotwell:
Thanks!

PR #646, opened 2023-02-17
[1.x] Use Application interface on ApplicationFactory

rogervila:
Relying on the Application contract provides flexibility when customizing a Laravel project.

The reason for that customization is explained here: https://github.com/laravel/framework/pull/46053#issuecomment-1424110252


rogervila:
I will reopen it for 2.x then.

PR #650, opened 2023-02-26
Fix timer table set failed

### LINKED ISSUE #648: Swoole\Table::set(): failed to set('161'), unable to allocate memory Due to fixed table size
- Octane Version: 1.4.0
- Laravel Version: 8.83.27
- PHP Version: 8.1.15
- Server & Version: Swoole 4.8.12

### Description:
I got the exception recently in the `php-fpm` log:

```php
 PHP Fatal error:  Uncaught ErrorException: Swoole\Table::set(): failed to set('161'), unable to allocate memory in /var/www/vendor/laravel/octane/src/Tables/SwooleTable.php:88
```
My server still has `10+` GB RAM, so it should not lack memory.
Then I found the tableSize in the `vendor/laravel/octane/bin/createSwooleTimerTable.php` is a const: `250`, 
```php
if (($serverState['octaneConfig']['max_execution_time'] ?? 0) > 0) {
    $timerTable = TableFactory::make(250);

    $timerTable->column('worker_pid', Table::TYPE_INT);
    $timerTable->column('time', Table::TYPE_INT);
    $timerTable->column('fd', Table::TYPE_INT);

    $timerTable->create();

    return $timerTable;
}
```
but I specified the worker number in the `octane:start` command is  `300`,  `php /var/www/artisan octane:start --host=0.0.0.0 --port=9000 --task-workers=600 --workers=300`

so I think I got this error because my worker number is larger than the const number `250`.

### Expected
Can tableSize be configurable or increased by worker number?

### Steps To Reproduce:
I've tried these steps on a local machine but I can't reproduce it.

1.  change `250` to `5` in `vendor/laravel/octane/bin/createSwooleTimerTable.php` 
2.  add a web route in the `routes/web.php`
``` php
Route::get('sleep', function(){
    sleep(5);
    return 'good';
});
```
3. start octane using command ` php artisan octane:start --workers=10 --max-requests=5`
4. send concurrent requests to route `sleep`
```php
Http::pool(function(Illuminate\Http\Client\Pool $pool){                                                                                                              array_map(
    fn($item) => $pool->get('http://127.0.0.1:8000/sleep'), 
    range(1,10));   
})
```

sy-records:
Fix #648 

The maximum number of rows a table can store is positively related to $size, but not exactly the same, e.g. $size is 1024 and the actual number of rows that can be stored is less than 1024.

So add a configuration item and let the user go and configure it according to the actual situation.

This code can be used to reproduce the problem.

```php
use Swoole\Table;

$table = new Table(1024);
$table->column('worker_pid', Table::TYPE_INT);
$table->column('time', Table::TYPE_INT);
$table->column('fd', Table::TYPE_INT);

$table->create();

foreach (range(1, 1050) as $i) {
    $table->set($i, [
        'worker_pid' => $i,
        'time' => time(),
        'fd' => $i,
    ]);
}

var_dump($table->stats(), $table->getSize());
```

### DIFF
--- a/bin/createSwooleTimerTable.php
+++ b/bin/createSwooleTimerTable.php
@@ -6,7 +6,7 @@ use Swoole\Table;
 require_once __DIR__.'/../src/Tables/TableFactory.php';
 
 if (($serverState['octaneConfig']['max_execution_time'] ?? 0) > 0) {
-    $timerTable = TableFactory::make(250);
+    $timerTable = TableFactory::make($serverState['octaneConfig']['max_timer_table_size'] ?? 250);
 
     $timerTable->column('worker_pid', Table::TYPE_INT);
     $timerTable->column('time', Table::TYPE_INT);
--- a/config/octane.php
+++ b/config/octane.php
@@ -218,4 +218,17 @@ return [
 
     'max_execution_time' => 30,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Maximum Timer Table Size
+    |--------------------------------------------------------------------------
+    |
+    | When using Swoole and max_execution_time is configured,
+    | it needs to be set according to the actual situation.
+    | The maximum number of rows a Table can store is positively related
+    | to this configuration item, but not identical.
+    |
+    */
+
+    'max_timer_table_size' => 250,
 ];


driesvints (approved):
Thanks @sy-records

PR #652, opened 2023-03-02
Add support for custom HTTP headers on static files

Starfox64:
This PR will enable users to add custom HTTP headers on files served from the public directory. This is important for production applications that could benefit from configuring a `Cache-Control` policy for their web assets. I have included an example policy to show how the new configuration option can be used.

This change should be backwards compatible as it checks if the new configuration option is present in `config/octane.php`.

### DIFF
--- a/config/octane.php
+++ b/config/octane.php
@@ -137,6 +137,21 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Static file headers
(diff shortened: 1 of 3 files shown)


driesvints:
It seems this breaks the test suite. Please resubmit this with passing tests.

PR #653, opened 2023-03-02
Add support for custom HTTP headers on static files

Starfox64:
This PR will enable users to add custom HTTP headers on files served from the public directory. This is important for production applications that could benefit from configuring a `Cache-Control` policy for their web assets. I have included an example policy to show how the new configuration option can be used.

This change should be backwards compatible as it checks if the new configuration option is present in `config/octane.php`.

### DIFF
--- a/config/octane.php
+++ b/config/octane.php
@@ -142,12 +142,14 @@ return [
     |
     | While using Swoole, you may serve static files from the public directory.
     | You may set the HTTP headers that should be returned along with those
-    | files here.
+    | files here. Headers can be applied according to the request path pattern.
     |
     */
 
     'static_file_headers' => [
-        // 'Cache-Control' => 'max-age=604800, must-revalidate, immutable, public',
+        // 'css/dist*' => [
+        //     'Cache-Control' => 'max-age=604800, must-revalidate, immutable, public',
+        // ],
     ],
 
     /*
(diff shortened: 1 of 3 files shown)


taylorotwell:
What if you want different headers for different files?

marden:
> What if you want different headers for different files?

Yes, totally agree. I have real cases when different headers are needed.

Starfox64:
I added the ability to apply the headers based on a request path pattern. This is the same pattern system that is used by some of the base middlewares.

taylorotwell:
So there is no way to also add support for this in RoadRunner?

Starfox64:
Unless I'm missing something, it seems the RoadRunner implementation doesn't even serve static files in the first place.

sneycampos:
Sorry for open a discussion on a merged PR but is not this in RoadRunner? https://roadrunner.dev/docs/middleware-static/2.x/en

sneycampos:
@Starfox64 sorry for open a discussion on a merged PR but is not this in RoadRunner? https://roadrunner.dev/docs/middleware-static/2.x/en

Starfox64:
> @Starfox64 sorry for open a discussion on a merged PR but is not this in RoadRunner? https://roadrunner.dev/docs/middleware-static/2.x/en

I guess the best way would be to just start Octane with `--rr-config= : The path to the RoadRunner .rr.yaml file` and use your own config file. Seems a lot more flexible than trying to wrap RR's config with Octane's.

PR #660, opened 2023-03-14
Add static file cache settings

jeffersonGlemos:
Add static file cache settings by @Starfox64


taylorotwell:
I think since we instruct people to serve Octane applications behind a webserver like Nginx or Apache, I will let users opt-in to this configuration option since I think it could be confusing. In production environments following our documentation, Swoole will not be serving any public assets.

https://laravel.com/docs/10.x/octane#serving-your-application-via-nginx

PR #663, opened 2023-03-21
This PR is here to run Laravel Octane CI against #46533 framework PR

vaawebdev:
https://github.com/laravel/framework/pull/46533

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,8 +16,8 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: ['8.0', 8.1, 8.2]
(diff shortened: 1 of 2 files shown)


vaawebdev:
@driesvints I had to replace the `laravel/framework` within the `repositories` config in `composer.json` to point to the last PR commit hash, define a duplicate package metadata & defined arbitrary `v10.4.10` framework version to avoid dependency conflicts. It seems that the change has no affect to laravel/octane

driesvints:
Thanks @vaawebdev. I'm not sure we can 100% rely on this as it might not be covered by a test. We'd need to test this in a real app.

PR #667, opened 2023-03-25
[1.x] Give new app instance to cache manager

smortexa:
Related PR: laravel/framework#46594


axlon (commented):
@smortexa you should add a `method_exists()` check, because octane supports Laravel 8, 9 and versions of 10 that don't have the method you added.

PR #672, opened 2023-04-04
[1.x] Fix RoadRunner binary windows

naquma:
Fixing first time `octane:install` command error on windows os


taylorotwell:
Octane does not support Windows.

PR #677, opened 2023-04-05
Exclude dependency from Node and Chokidar for RR

masterjus:
The RoadRunner has the [debug option](https://github.com/roadrunner-server/roadrunner/blob/master/.rr.yaml#L836) that allows to reload worker after each request. That means that you don't need Node and Chokidar library to be installed for development.


taylorotwell:

masterjus:
@taylorotwell 
This is not a bug fix but more correct feature implementation.
As I mentioned in the PR description, my fix removes requirement to install the package chokidar for RoadRunner users. The functionality stays the same.

masterjus:
@nunomaduro What do you think about this change?

frans-slabbekoorn:
Have you had any time to look at this request @taylorotwell?

PR #679, opened 2023-04-18
add ServerProcessInspector interface

yiming0:


yiming0:
> Please provide a thorough explanation why you need this.

While developing a solution to start `Laravel Octane` using other PHP asynchronous frameworks, I encountered a warning about the parameter type hinting in `Laravel\Octane\Commands\Concerns\InteractsWithServers::runServer`.

I think `Laravel Octane` will not only serve `swoole` and `roadrunner` in the future, so I have added the `\Laravel\Octane\Contracts\ServerProcessInspector` interface.

PR #682, opened 2023-04-21
Use Laravel ResponseHeaderBag to send Set-Cookie Header instead of Swoole function

NiroDeveloper:
At the moment the SwooleClient uses the cookie or rawCookie function of the SwooleResponse to send cookies.
With these functions it is not possible to send the Max-Age cookie parameter, only an absolute expires time is sent.
Which leads to the problem that clients without a valid time do not save the cookie correctly.

To fix this and simplify the code, this pull request removes the swoole cookie function and reuses the cookie header function from the Laravel ResponseHeaderBag class.

This has the further advantage that the responses between a normal Laravel PHP-FPM server and Laravel Octane do not differ.

on-behalf-of: @e-solutions-GmbH <info@esolutions.de>

### DIFF
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -171,8 +171,9 @@ class SwooleClient implements Client, ServesStaticFiles
     /**
      * Send the headers from the Illuminate response to the Swoole response.
      *
-     * @param  \Symfony\Component\HttpFoundation\Response  $response
-     * @param  \Swoole\Http\Response  $response
+     * @param   \Symfony\Component\HttpFoundation\Response  $response
+     * @param   \Swoole\Http\Response                       $swooleResponse
+     *
      * @return void
      */
     public function sendResponseHeaders(Response $response, SwooleResponse $swooleResponse): void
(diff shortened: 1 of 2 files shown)


src/Swoole/SwooleClient.php
```diff
@@ -171,8 +171,9 @@ public function respond(RequestContext $context, OctaneResponse $octaneResponse)
     /**
      * Send the headers from the Illuminate response to the Swoole response.
      *
-     * @param  \Symfony\Component\HttpFoundation\Response  $response
-     * @param  \Swoole\Http\Response  $response
+     * @param   \Symfony\Component\HttpFoundation\Response  $response
+     * @param   \Swoole\Http\Response                       $swooleResponse
+     *
```
driesvints, line 176: Please undo these changes.

NiroDeveloper: Reverted now ... sorry, problems with IDE formatting

taylorotwell:
Eh, I'm a little uncomfortable just deleting this code. It was put there for a reason. We would need better explanation and more confidence from a Swoole maintainer that just deleting this code is the proper answer.

NathanFreeman:
@NiroDeveloper @taylorotwell @driesvints Hi.
I have made a pr about this problem and maybe it would be helpful.
https://github.com/swoole/swoole-src/pull/5034
https://github.com/swoole/swoole-src/pull/5035

PR #691, opened 2023-05-03
feat: enable http compression by sending data not with chunked encodi…

DeepDiver1975:
…ng - if possible

http compression can only be used if chunked encoding is not used.
Therefore if the response body "fits into one chunk" `end()` is called instead of `write()` followed by `end()`

refs: https://openswoole.com/docs/modules/swoole-http-server/configuration#http_compression

Applies to swoole as well

### DIFF
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -260,12 +260,15 @@ class SwooleClient implements Client, ServesStaticFiles
             return;
         }
 
+        // if the content fits into one chunk -> perform no chunked encoding
         if ($length <= $this->chunkSize) {
-            $swooleResponse->write($content);
-        } else {
-            for ($offset = 0; $offset < $length; $offset += $this->chunkSize) {
-                $swooleResponse->write(substr($content, $offset, $this->chunkSize));
-            }
+            $swooleResponse->end($content);
+
+            return;
+        }
+
+        for ($offset = 0; $offset < $length; $offset += $this->chunkSize) {
+            $swooleResponse->write(substr($content, $offset, $this->chunkSize));
         }
 
         $swooleResponse->end();
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -260,7 +260,6 @@ class SwooleClient implements Client, ServesStaticFiles
             return;
         }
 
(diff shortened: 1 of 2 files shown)


DeepDiver1975:
> Can you add tests on this please?

Two tests have been added:
- one to test the chunked behavior
- one to test the not chunked behavior

taylorotwell:
Has conflicts.

taylorotwell:
Please mark as ready for review when fixed.

DeepDiver1975:
@taylorotwell conflict resolved - thx

DeepDiver1975:
THX @taylorotwell

alexryall:
This change contains a breaking change for us. Currently we are GZIP encoding responses with the below code and with this change the server returns malformed responses. For now I've locked our Octane version to 1.5.4, what would be the recommended way to GZIP encode responses? I don't want to add an extra layer e.g. Nginx just to encode the response.

`<?php

namespace App\Http\Middleware;

use Closure;
use Illuminate\Http\Request;

class GzipEncodeResponse
{
    /**
     * @param Request $request
     * @param Closure $next
     * @return \Illuminate\Http\Response|\Illuminate\Http\JsonResponse
     */
    public function handle(Request $request, Closure $next)
    {
        $response = $next($request);

        if (in_array('gzip', $request->getEncodings()) && function_exists('gzencode')) {
            $response->setContent(gzencode($response->getContent(), 5));
            $response->headers->add([
                'Content-Encoding' => 'gzip',
            ]);
        }

        return $response;
    }
}
`

alexryall:
@DeepDiver1975 thanks for explaining, that makes sense. I did try to get the automatic compression working with the official Swoole docker image (using [this](https://hub.docker.com/layers/phpswoole/swoole/5.0.3-php8.2/images/sha256-8a63c5dff3266314c03eef25999b98ec17bb10350f25071835c660d9c99e9453?context=explore) one currently) but didn't have any joy so ended up doing it within Laravel. I thought it was probably a combination of http_compression not being set to true within Octane and the way the Docker image has been setup.

alexryall:
With a bit of messing around managed to get Brotli and Gzip working thanks to this change. You're right about the correct packages (libz/brotli) needing to be installed before Swoole is installed.

PR #700, opened 2023-05-25
Disable swoole header name formatting by config value

NiroDeveloper:
Swoole and Open Swoole are formatting headers by default, but it can be disabled.
https://openswoole.com/docs/modules/swoole-http-response-header

In the SwooleClient class, the header will be read with the original formatting.
$headers = $response->headers->allPreserveCase();

But if you give these headers to
$swooleResponse->header($name, $value);
Swoole will format the header because the third parameter $format is not set.

The swoole formatting should be adjustable by config.
(I know I should use rfc compliant headers, but i can not update old weird clients)


driesvints:
I'm not sure why I'm author or one of these commits? Please resubmit this and remove me as an author of these commits.

PR #702, opened 2023-05-26
Disable swoole header name formatting by config value

NiroDeveloper:
Swoole and Open Swoole are formatting headers by default, but it can be disabled.
https://openswoole.com/docs/modules/swoole-http-response-header

In the SwooleClient class, the header will be read with the original formatting.
$headers = $response->headers->allPreserveCase();

But if you give these headers to
$swooleResponse->header($name, $value);
Swoole will format the header because the third parameter $format is not set.

The swoole formatting should be adjustable by config.
(I know I should use rfc compliant headers, but i can not update old weird clients)

This change is backwards compatible, because the default is still "true".

on-behalf-of: @e-solutions-GmbH <info@esolutions.de>

### DIFF
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -119,10 +119,12 @@ class SwooleClient implements Client, ServesStaticFiles
         $octaneConfig = $context->octaneConfig ?? [];
 
         if (! empty($octaneConfig['static_file_headers'])) {
+            $headerNameFormatting = config('octane.swoole.header_name_formatting', true);
+
             foreach ($octaneConfig['static_file_headers'] as $pattern => $headers) {
                 if ($request->is($pattern)) {
                     foreach ($headers as $name => $value) {
-                        $swooleResponse->header($name, $value);
+                        $swooleResponse->header($name, $value, $headerNameFormatting);
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


taylorotwell:

NiroDeveloper:
@taylorotwell 
I understand your concerns, but for my use-case this is clearly a bug which is stopping me from using Octane.
As already mentioned, i have to serve clients who are already a few years old and can not be updated.

Unfortunately, these were implemented badly, the headers are hardcoded and case-sensitive.
Because Swoole is now formatting my input, it can not read the header values and is crashing.

Harti:
It seems odd that you specifically called ResponseHeaderBag's allPreserveCase presumably to... preserve the casing... 
https://github.com/laravel/octane/blob/d962806a79c5ac97ddd1e371bd232d00c855c0d9/src/Swoole/SwooleClient.php#L156

... but are suggesting it's fine for Swoole to override the casing few lines later? 
https://github.com/laravel/octane/blame/d962806a79c5ac97ddd1e371bd232d00c855c0d9/src/Swoole/SwooleClient.php#L161

To be fair, Swoole's API probably didn't expose that option back then - but how is an optional toggle not considered a desired bugfix here?

Harti:
@taylorotwell What do you mean by delaying the merge - will this PR get looked at in the future as an improvement, and what's the rough release window for that? 

FWIW, it should be considered a bug: Both Symfony and Illuminate preserve the header casing, and it works if you use `artisan serve`. Once you use Octane, though, it's no longer preserved - i.e. **Octane is breaking something** that's expected to work.

This is a non-breaking PR, keeping defaults at the current behavior. Is there any risk that I may be missing?

PR #705, opened 2023-06-07
RoadRunner bridge for laravel cache and config improvements

asanikovich:
I offer to add the RoadRunner bridge for Laravel Cache. It will work by RPC.

Also, I prepare some config improvements to correctly set up rr server.
I am ready to update related docs and add more tests if you say I am on the right path.

Changes:
To install RoadRunner we require package "spiral/roadrunner-kv".

### DIFF
--- a/.github/workflows/tests-rr.yml
+++ b/.github/workflows/tests-rr.yml
@@ -1,0 +1,59 @@
+name: tests
+
+on:
+  push:
+    branches:
+      - master
+      - '*.x'
+  pull_request:
+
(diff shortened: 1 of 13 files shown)


taylorotwell:
Could this be published as a package?

asanikovich:
@taylorotwell 
I make a package with full coverage (tested in a real app).
https://github.com/asanikovich/laravel-roadrunner-cache
Can you promote it somewhere in docs related to octane?

FYI @smortexa

Issue #545, opened 2022-06-22
Swoole\Http\Response::create(): parameter $2 must be valid connection session id


ibrunotome:
- Octane Version: v1.2.12
- Laravel Version: v9.18.0
- PHP Version: 8.1.0
- Server & Version: Swoole / RoadRunner openswoole/swoole:4.11.1-php8.1-alpine
- Database Driver & Version: postgres 14.1-alpine

### Description:

Debuging in a semi-production environment with cloudflare pointing to google cloud load balancer IP -> GKE ingress -> app:8000, a login request works like 14/15 of the times, but that 1/15:

<img width="1444" alt="Screen Shot 2022-06-21 at 21 41 38" src="https://user-images.githubusercontent.com/4256471/174920103-3cfed8ea-9035-4579-93e6-9bd5923434f8.png">

Any thoughts about it? I know about the docs recommendation to run octane behind a nginx server, but I don't know if there is any correlation, since it's a layer on top of it which will call the same endpoint.

ibrunotome:
more info:

FortifyServiceProvider:

```php
class FortifyServiceProvider extends ServiceProvider
{
    public function boot(): void
    {
        Fortify::authenticateUsing([new AuthenticateUser(), '__invoke']);
    }
}
```

my login action

```php
<?php

namespace App\Actions\Fortify;

use App\Enums\Ability;
use App\Models\User;
use Illuminate\Database\Eloquent\Model;
use Illuminate\Support\Facades\Hash;
use Illuminate\Validation\ValidationException;
use Laravel\Fortify\Fortify;
use Laravel\Fortify\Http\Requests\LoginRequest;

class AuthenticateUser
{
    /**
     * @throws \Illuminate\Validation\ValidationException
     */
    public function __invoke(LoginRequest $request): Model|User
    {
        $validated = $request->validated();
        $validated[Fortify::username()] = mb_strtolower($validated[Fortify::username()]);

        $user = User::where('email', '=', $validated[Fortify::username()])
            ->orWhere('nickname', '=', $validated[Fortify::username()])
            ->orWhere('phone_number', '=', $validated[Fortify::username()])
            ->first();

        if (null !== $request->session()->get('back_to')) {
            $backTo = $request->session()->get('back_to');
            session()->forget('back_to');
            config(['fortify.redirects.login' => $backTo]);
        }

        if ($user && Hash::check($request->input('password'), $user->password)) {
            if (!$user->can(Ability::Login->value)) {
[...]

Issue #712, opened 2023-06-28
Incompatibility Issue with Firestore Leading to Mutex Error and Invalid Connection Session ID


Jakoffe:
### Octane Version

2.0.1

### Laravel Version

10.14.0

### PHP Version

8.2.4

### What server type are you using?

Swoole

### Server Version

5.0.3

### Database Driver & Version

MySQL 8.0.31 on both amd64 and arm64

### Description

Hello Maintainers and Contributors,

First, I would like to extend my gratitude for all the effort put into Laravel Octane. I have encountered an issue while using Laravel Octane with Swoole and Google's Firestore package for in-app messaging. I would like to know if anyone can provide insights or solutions to this problem. The problem occurs in our development, staging, and production environments.

**We use the following Firestore related packages:**
- "google/cloud-firestore": "1.35.0"
- "grpc/grpc": "1.52.0"
- "kreait/laravel-firebase": "5.2.0"

**Issue**
When attempting to utilize the chat functionality in our applications (flutter app and admin-panel), which are built using Firestore, we receive the following error message in the logs:

`[mutex.cc : 2417] RAW: Check w->waitp->cond == nullptr failed: Mutex::Fer while waiting on Condition`

This error seems to originate from a low-level C++ library, possibly the underlying gRPC library used by Firestore. Shortly after this error, we also receive the following warning message:

`Warning: Swoole\Http\Response::create(): parameter $2 must be valid connection session id in /var/www/vendor/laravel/octane/src/Swoole/Actions/EnsureRequestsDontExceedMaxExecutionTime.php on line 33`

This error appears to be a side effect of the first error, indicating that the connection might have gotten into a bad state due to the underlying library issue.

Every other aspect of our backend works great with Octane (Swoole).

**What was tried:**
- I have tried updating all dependencies to the latest versions.
- I have checked the configuration of both Firestore and Laravel Octane.
- I have attempted to isolate the issue by using Firestore without Octane/Swoole and vice versa (Before moving to Octane we have been running the backend using php-fpm for 12+ months without any issues).

### Steps To Reproduce

```php
namespace App\Models\Firestore;

use App\Models\Staff;
use App\Models\User;
use Exception;
use Google\Cloud\Core\Timestamp;
use Google\Cloud\Firestore\CollectionReference;
use Google\Cloud\Firestore\DocumentReference;
use Google\Cloud\Firestore\FieldValue;
use Google\Cloud\Firestore\FieldValue\ArrayUnionValue;
[...]

PR #737, opened 2023-08-25
Fix passing invalid connection session id to `Swoole\Http\Response::create()`

smortexa:
Fixes laravel/octane#545 and potential fix for laravel/octane#712

### DIFF
--- a/src/Swoole/Actions/EnsureRequestsDontExceedMaxExecutionTime.php
+++ b/src/Swoole/Actions/EnsureRequestsDontExceedMaxExecutionTime.php
@@ -27,6 +27,10 @@ class EnsureRequestsDontExceedMaxExecutionTime
             if ((time() - $row['time']) > $this->maxExecutionTime) {
                 $this->timerTable->del($workerId);
 
+                if ($this->server instanceof Server && ! $this->server->exists($row['fd'])) {
+                    continue;
+                }
+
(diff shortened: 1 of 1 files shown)


src/Swoole/Actions/EnsureRequestsDontExceedMaxExecutionTime.php
```diff
@@ -27,6 +27,10 @@ public function __invoke()
             if ((time() - $row['time']) > $this->maxExecutionTime) {
                 $this->timerTable->del($workerId);
 
+                if ($this->server instanceof Server && ! $this->server->exists($row['fd'])) {
```
sy-records, line 30: Looks like a good patch. But what is the reproduction code? 

When I test with this code, I close the browser and `exists` still return `true`.

```php
$http = new Swoole\Http\Server('0.0.0.0', 9502);

$http->on('Request', function ($request, $response) use ($http) {
    sleep(10);
    var_dump($http->exists($request->fd));
    $response->end('Hello Swoole');
});

$http->start();
```

Can you add a test case?

smortexa: I encountered this issue only in a production environment and within a containerized setup. It seems that others who reported the issue are facing the same situation. However, despite my attempts, I was unable to reproduce this issue in a test case.

PR #742, opened 2023-08-28
Add “raw” type in handleStream method for custom json in stdout

mphamid:
To implement the filebeat logs in stdout during the execution of the project with octane, all outputs are displayed with the info method in the output, which causes a change in the output of stdout and makes it difficult to implement filebeat using stdout.
My suggestion is to create a new type in the handleStream method and use it whenever we need to send a specific json in the output.

### DIFF
--- a/src/Commands/Concerns/InteractsWithIO.php
+++ b/src/Commands/Concerns/InteractsWithIO.php
@@ -238,6 +238,7 @@ trait InteractsWithIO
             'request' => $this->requestInfo($stream, $verbosity),
             'throwable' => $this->throwableInfo($stream, $verbosity),
             'shutdown' => $this->shutdownInfo($stream, $verbosity),
+            'raw' => $this->raw(json_encode($stream)),
             default => $this->info(json_encode($stream), $verbosity)
         };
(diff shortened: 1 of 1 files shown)


taylorotwell:
How do you use this?

mphamid:
I created a new log channel
```php
'filebeat_stdout' => [
            'driver' => 'monolog',
            'handler' => StreamHandler::class,
            'formatter' => FilebeatFormatter::class,
            'level' => env('LOG_LEVEL', 'debug'),
            'with' => [
                'stream' => 'php://stdout',
            ],
        ],
```
And I created the output in FilebeatFormatter::class file like this
```php
class FilebeatFormatter extends JsonFormatter
{
    /**
     * @inheritDoc
     */
    public function format(LogRecord $record): string
    {
        $normalized = json_decode(parent::format($record),true);
        $normalized['application'] = config('app.name');
        $normalized['type'] = 'raw';
        return $this->toJson($normalized, true) . ($this->appendNewline ? "\n" : '');
    }
}
```
But now my output is showing an info in stdout

PR #764, opened 2023-10-26
[2.x] Add support for FrankenPHP

dunglas:
<p align="center"></p>

This patch adds [FrankenPHP](https://frankenphp.dev) as a runtime for Laravel Octane!

To get started, use the following commands:

```console
php artisan octane:install --server=frankenphp
php artisan octane:start
```

The current implementation provides feature parity with RoadRunner (worker mode, watch, and max requests) but also adds new features not supported by the other engines:

* Production-grade HTTPS (with automatic certificate creation and renewal), HTTP/2, and HTTP/3 (enabled only if the new `--https` flag is passed)
* Real-time updates with the built-in [Mercure](https://mercure.rocks) hub. Install [`mvanduijker/laravel-mercure-broadcaster`](https://github.com/mvanduijker/laravel-mercure-broadcaster) for seamless integration with Laravel Broadcast.
* Support for sending [103 HTTP Early Hints responses](https://github.com/dunglas/frankenphp/blob/main/docs/early-hints.md) (assets preloading), (see https://github.com/symfony/symfony/pull/48128 for usage examples)
* Automatic compression using Zstandard or Gzip, depending on client support


taylorotwell:
Very cool - will review! 🙏

driesvints:
Very cool @dunglas!

vasilvestre:
The description about the https isn't clear, feels like HTTP3 is enabled when https is on
Apart from it, incredible work :)

dunglas:
@vasilvestre thanks! To clarify: by default, only HTTP/1.1 (unencrypted) is enabled. When `--https` is passed, the connection is encrypted and HTTP versions 1.1, 2 and 3 are supported.

src/Commands/StatusCommand.php
```diff
@@ -65,6 +67,17 @@ protected function isRoadRunnerServerRunning()
             ->serverIsRunning();
     }
 
+    /**
+     * Check if the FrankenPHPO server is running.
```
williamdes, line 71:
```suggestion
     * Check if the FrankenPHP server is running.
```

roberto-butti:
super, i will try it!

src/Commands/stubs/Caddyfile
```diff
@@ -0,0 +1,38 @@
...
+
+	route {
+		encode zstd gzip
+
+		# Mercure config is injected here
+		{$CADDY_SERVER_EXTRA_DIRECTIVES}
+
+		# If the requested file does not exist, pass to the worker script
+		@indexFiles file {
+			try_files public/{path} vendor/laravel/octane/bin/frankenphp-worker.php
```
francislavoie, line 30: This is a security vulnerability.

The Caddyfile should have `root public/` to ensure the webroot is correctly set. While the `try_files` does rewrite with `public/` it's still a risk to have the webroot set incorrectly because a request path like `/../config/secrets.txt` could still escape out of `public/`. Caddy has special path handling to ensure cleaned paths don't escape the root, but if the root is not correctly set, then those protections aren't in place.

IIRC Laravel packages can publish files to userland directories. I think `frankenphp-worker.php` should probably be published to `public/`. That'll be necessary so that Caddy sees that file within the `public/` webroot instead of elsewhere. I haven't written Laravel packages in quite a while, so help wanted here to do this correctly. IIRC there's `public_path()` which can be used to get the correct directory to publish the file to.

src/Commands/stubs/Caddyfile
```diff
@@ -0,0 +1,38 @@
...
+			split_path .php
+		}
+		rewrite @indexFiles {http.matchers.file.relative}
+		@phpFiles path *.php
+		php @phpFiles
```
francislavoie, line 35: This can be replaced with `php_server`, no? Can do this:

```
php_server {
	try_files {path} {path}/ /frankenphp-worker.php
}
```

Assumes that the worker script is moved to the `public/` webroot.

francislavoie (commented):
:wave: Hey everyone! @dunglas and I spoke a bunch about FrankenPHP in the past; I'm a Caddy maintainer by night, and a PHP dev by day, so these comments come from that perspective.

config/octane.php
```diff
@@ -51,6 +51,29 @@
...
+    |
+    | This option configures the Mercure hub (events broadcasting) that is included in FrankenPHP.
+    | The defaults are good enough for most applications.
+    | "publisher_jwt" configures the key and algorithm to use for verifying JWTs provided by publishers.
+    | "subscriber_jwt" configures the key and algorithm to use for verifying JWTs provided by subscribers.
+    | Set "anonymous" to false to disable anonymous subscribing.
+    | Set "subscriptions" to false to disable the subscription API.
+    | Set "mercure" to false to entirely disable the Mercure hub.
+    |
+    | More information about Mercure can be found on https://mercure.rocks.
```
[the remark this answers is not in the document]
dunglas: I'm aware of that but I fear that it will make the linter red and as you pointed out, in this case the link is valid even with the final dot.

bin/frankenphp-worker.php
```diff
@@ -0,0 +1,43 @@
...
+                )
+            )->boot();
+
+            [$request, $context] = $frankenPhpClient->marshalRequest(new RequestContext());
+
+            $worker->handle($request, $context);
+        })
+    ) {
+        $nbRequests++;
+    }
```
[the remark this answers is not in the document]
dunglas: Good idea! We should also do that in the official documentation and in the worker script provided by the `php-runtime` project!

src/Commands/Concerns/InstallsFrankenPhpDependencies.php
```diff
@@ -0,0 +1,103 @@
...
+     */
+    protected function downloadFrankenPhpBinary()
+    {
+        $arch = php_uname('m');
+
+        $assetName = null;
+        if (PHP_OS_FAMILY === 'Linux') {
+            if ($arch !== 'x86_64') {
+                $assetName = 'frankenphp-linux-x86_64';
+            }
```
Namoshek, line 44: This needs a second look.

And I would suggest replacing the whole section with a `match` clause for readability:
```php
$assetName = match (true) {
    PHP_OS_FAMILY === 'Linux' && $arch === 'x86_64' => 'frankenphp-linux-x86_64',
    PHP_OS_FAMILY === 'Darwin' => "frankenphp-mac-$arch",
    default => null,
};
```

src/Commands/stubs/Caddyfile
dunglas: Unless I missed something, there are now vulnerabilities here.
We use `try_files` to check that the file exists in `public/`, and we use the hardcoded path to the worker script if it doesn't.

Try `curl http://127.0.0.1:8000/../package.json` for instance, and you get a 404, as expected.

Publishing the worker script in `public/` was indeed my first idea. FrankenPHP is designed to work like that. Ideally, the best option would even be to check if we're using FrankenPHP in worker mode or another server in `public/index.php`. But it's not how Octane is currently designed so I used this "hack" to make something a close as possible to the current design.

If the maintainers agree, Im totally ok to switch to publishing the script in `public/` (it would more fit FrankenPHP's design and will allow removing these hacks in the Caddyfile). WDYT @taylorotwell @driesvints?

src/Commands/stubs/Caddyfile
dunglas: Currently, we can't (see https://github.com/laravel/octane/pull/764#discussion_r1375212084), but if we change the design to publish the worker script in `public/` then indeed we'll be able to simplify the config.

src/Commands/Concerns/InstallsFrankenPhpDependencies.php
dunglas: Good idea!

src/Commands/stubs/Caddyfile
dunglas: I refactored the code to publish the worker script. It's much simpler now and the vuln is gone!

src/Commands/stubs/Caddyfile
dunglas: done

src/Commands/InstallCommand.php
```diff
@@ -116,6 +118,29 @@ public function installSwooleServer()
...
+                        File::append(
+                            base_path('.gitignore'),
+                            $file.PHP_EOL
+                        );
+                    }
+                });
+        }
+
+        return $this->ensureFrankenPhpBinaryIsInstalled();
+    }
```
KennedyTedesco, line 142: The `installFrankenPhpServer()` method works well, but I'd prefer a version with fewer file operations, even though it's not a big deal in the context of this part (installation, etc.).

My version (of course, you can completely ignore it if you want :blush: ):

```suggestion
    public function installFrankenPhpServer()
    {
        $gitIgnorePath = base_path('.gitignore');

        if (File::exists($gitIgnorePath)) {
            $contents = File::get($gitIgnorePath);

            $filesToAppend = collect(['frankenphp', 'Caddyfile', 'frankenphp-worker.php'])
                ->filter(function ($file) use ($contents) {
                    return !str_contains($contents, $file . PHP_EOL);
                })
                ->implode(PHP_EOL);

            if ($filesToAppend !== '') {
                File::append($gitIgnorePath, PHP_EOL . $filesToAppend . PHP_EOL);
            }
        }

        return $this->ensureFrankenPhpBinaryIsInstalled();
    }
```

francislavoie: A couple more lines can be saved with:

```php
                ->filter(fn ($file) => !str_contains($contents, $file . PHP_EOL))
```

src/Commands/StartFrankenPhpCommand.php
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    /**
+     * Get the number of workers that should be started.
+     *
+     * @return int
+     */
+    protected function workerCount()
+    {
+        return $this->option('workers') == 'auto'
```
KennedyTedesco, line 99:
```suggestion
        return $this->option('workers') === 'auto'
```

src/Commands/InstallCommand.php
KennedyTedesco: > A couple more lines can be saved with:
> 
> ```
>                 ->filter(fn ($file) => !str_contains($contents, $file . PHP_EOL))
> ```

Suggestion applied. 👌

src/Commands/stubs/frankenphp-worker.php
```diff
@@ -0,0 +1,50 @@
...
+    };
+    while (
+        $nbRequests < ($_ENV['MAX_REQUESTS'] ?? $_SERVER['MAX_REQUESTS']) &&
+        frankenphp_handle_request($handleRequest)
+    ) {
+        $nbRequests++;
+    }
+} finally {
+    $worker?->terminate();
+}
```
KennedyTedesco, line 50: Do we need to perform this operation on every iteration?

```
($_ENV['MAX_REQUESTS'] ?? $_SERVER['MAX_REQUESTS'])
```

In this version I have extracted this to a `$maxRequests` variable. Does it make sense or I missed something? If I missed, sorry. =D

```suggestion
$frankenPhpClient = new FrankenPhpClient();

$worker = null;
$nbRequests = 0;
$maxRequests = $_ENV['MAX_REQUESTS'] ?? $_SERVER['MAX_REQUESTS'];
try {
    $handleRequest = static function () use (&$worker, $basePath, $frankenPhpClient) {
        $worker ??= tap(
            new Worker(
                new ApplicationFactory($basePath), $frankenPhpClient
            )
        )->boot();

        [$request, $context] = $frankenPhpClient->marshalRequest(new RequestContext());

        $worker->handle($request, $context);
    };
    while (
        $nbRequests < $maxRequests &&
        frankenphp_handle_request($handleRequest)
    ) {
        $nbRequests++;
    }
} finally {
    $worker?->terminate();
}
```

src/Commands/InstallCommand.php
dunglas: Thanks! I must admit that I copy/pasted/adapted code from other engines but I prefer your approach too!

src/FrankenPhp/Concerns/FindsFrankenPhpBinary.php
```diff
@@ -0,0 +1,27 @@
...
+trait FindsFrankenPhpBinary
+{
+    /**
+     * Find the FrankenPHP binary used by the application.
+     */
+    protected function findFrankenPhpBinary(): ?string
+    {
+        if (file_exists($basePath = base_path('frankenphp'))) {
+            return $basePath;
+        }
```
KennedyTedesco, line 17: This part makes me wonder if we should check whether it's actually a file and an executable one, like this:

```suggestion
        $frankenPhpBinary = base_path('frankenphp');

        if (is_file($frankenPhpBinary) && is_executable($frankenPhpBinary)) {
            return $frankenPhpBinary;
        }
```

Because `file_exists()` could be anything, like an empty text file or a binary without execution privileges or even a folder called `/frankenphp`, and so on.

Maybe I'm being too strict here, so, you can just ignore this. =D

dunglas: Actually, this block is useless (it was a copy/paste from the RoadRunner installer, which also contains useless code). Fixed.

KennedyTedesco: @dunglas I think the idea behind the old code was: "Let's prioritize finding the binary in base_path. As a last resort, we'll try to find it in the `$PATH`." By doing this: `(new ExecutableFinder())->find('frankenphp', null, [base_path()]);`, I'm pretty confident that it will first try to find the binary in various $PATH and finally in base_path as the last one. 

But I don't think this is an issue, right? I think it's okay. =)

Magiczny375:
Cool man, I'll try this!

config/octane.php
```diff
@@ -51,6 +51,29 @@
...
+    */
+
+    'mercure' => [
+        'publisher_jwt' => env('MERCURE_SUBSCRIBER_JWT_KEY', '!ChangeThisMercureHubJWTSecretKey!'),
```
shalvah, line 71: Shouldn't this be

```suggestion
        'publisher_jwt' => env('MERCURE_PUBLISHER_JWT_KEY', '!ChangeThisMercureHubJWTSecretKey!'),
```

dunglas: Yes, good catch!

taylorotwell:
Drafting pending review from @nunomaduro

jihadismail8:
@dunglas does Franken supports GRPC(php as client at least ) just asking because RoadRunner does ?

jihadismail8:
@dunglas thanks for your answer, I am really interested in franken . I hope there is gona be a little bit more docs about it. I just tried it on an already existing laravel project by ( docker run -p 443:443 -v $PWD:/app dunglas/frankenphp ) but somehow getting an error of connection to my postgres database ( which is running on the host machine , not as a separated docker container) tried in env. To use localhost, 0.0.0.0 , 127.0.0.1 , docker.host.local . And none worked.  Strange

francislavoie:
@jihadismail8 That doesn't have anything to do with PHP/Laravel/FrankenPHP, that's just how Docker works. You need to add `--add-host=host.docker.internal:host-gateway` when you run your container, then you can use `host.docker.internal:5432` to connect to your postgres DB on the host. Or, run postgres in a Docker container in the same network as frankenphp, then use the container name to connect to it.

dunglas:
@fannyfan414 the main advantage is simplicity: fewer moving parts, only one binary, no network or UNIX socket connection between the web server and the application server (it's the same process), etc.

Regarding performance, there is currently a problem with this PR: @nunomaduro detected that it's currently slow (an order of magnitude slower than Symfony and other frameworks in worker mode, and even than Laravel without worker mode). I'll investigate and do some benchmarks when this will be fixed.

nunomaduro:
Hi @dunglas, I've just retested this pull request, and it seems something is still not right. Here's a basic load test comparing Swoole and Franken PHP:

```
Use case: basic Laravel application, with Breeze/Blade/Alpine:
-- ./vendor/bin/pest stress http://127.0.0.1:8000
  Swoole:
  Test Duration ....................................................... 5.00 s
  Test Concurrency ......................................................... 1
  Requests Count .................................. 167.71 reqs/s 839 requests
  Request Duration ................................................... 5.27 ms
  
  Franken PHP:
  Test Duration ....................................................... 5.20 s
  Test Concurrency ......................................................... 1
  Requests Count ..................................... 3.84 reqs/s 20 requests
  Request Duration ................................................. 326.23 ms
  
-- ./vendor/bin/pest stress http://127.0.0.1:8000 --concurrency=8
  
  Swoole:
  Test Duration ....................................................... 5.10 s
  Test Concurrency ......................................................... 8
  Requests Count ................................. 801.34 reqs/s 4083 requests
  Request Duration ................................................... 7.24 ms

  Franken PHP:
  Test Duration ....................................................... 7.37 s
  Test Concurrency ......................................................... 8
  Requests Count ..................................... 3.39 reqs/s 25 requests
  Request Duration ................................................ 2514.31 ms
```

So, at the moment, this basic load test concludes the following:

```
1. Franken PHP 43x slower than Swoole with 1 concurrency
2. Franken PHP 236x slower than Swoole with 8 concurrency
```

@dunglas When you were benchmarking Swoole/Roadrunner versus FrankenPHP on Symfony, were you able to achieve the same or better levels of performance?

dunglas:
For Symfony, we have benchmarks comparing FrankenPHP to FPM but not to RoadRunner/Swoole as they don't have the same purpose (the main goal of FrankenPHP is to simplify deployment because there is only 1 service, automatic HTTPS, native HTTP/2 and 3...).

That being said, the difference looks important. I'll try to see what are the difference and if it can be improved.

Disabling access logs, tweaking the number of workers and disabling the Go GC (which is known to not be well suited for stress tests, but is generally not an issue in prod) usually helps to improve performance a lot.

nunomaduro:
@dunglas Thank you for time on this one.

dunglas:
Here is what I get with the latest version of this PR and of FrankenPHP (default Laravel Homepage, Breeze with Blade installed).

Gzip/Zstandard disabled:

```
./vendor/bin/pest stress http://127.0.0.1:8000      

  Stress testing 127.0.0.1:8000 ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― for 5 seconds  
  00:25:40                                                                                                                                    0.90ms  
  00:25:41                                                                                                                                    0.93ms  
  00:25:42                                                                                                                                    0.93ms  
  00:25:43                                                                                                                                    0.93ms  
  00:25:44                                                                                                                                    0.94ms  
  Result ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――  
  Test Duration ............................................................................................................................. 5.00 s  
  Test Concurrency ............................................................................................................................... 1  
  Requests Count ....................................................................................................... 659.91 reqs/s 3300 requests  
  Success Rate ............................................................................................................................. 100.0 %  
  DNS Lookup Duration .....................................................................................................................  0.00 ms  
  TLS Handshake Duration ................................................................................................................... 0.00 ms  
  Request Duration ......................................................................................................................... 0.94 ms  
  — Upload 0.3 % ..................................................................................................... 0.00 MB/req 0.07 MB/s 0.00 ms  
  — TTFB 95.2 % — including server processing time ......................................................................................... 0.89 ms  
  — Download 4.4 % .................................................................................................. 0.00 MB/req 12.10 MB/s 0.04 ms  
  ■ Critical  ■ Poor  ■ Ok  ■ Excellent  
```

```
➜  frankenphp-laravel ./vendor/bin/pest stress http://127.0.0.1:8000 -concurrency=8

  Stress testing 127.0.0.1:8000 ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― for 5 seconds  
  00:26:16                                                                                                                                    0.99ms  
  00:26:17                                                                                                                                    0.99ms  
  00:26:18                                                                                                                                    0.99ms  
  00:26:19                                                                                                                                    0.98ms  
  Result ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――  
  Test Duration ............................................................................................................................. 5.01 s  
  Test Concurrency ............................................................................................................................... 1  
  Requests Count ....................................................................................................... 453.24 reqs/s 2272 requests  
  Success Rate ............................................................................................................................. 100.0 %  
  DNS Lookup Duration .....................................................................................................................  0.00 ms  
  TLS Handshake Duration ................................................................................................................... 0.00 ms  
  Request Duration ......................................................................................................................... 0.99 ms  
  — Upload 0.3 % ..................................................................................................... 0.00 MB/req 0.05 MB/s 0.00 ms  
  — TTFB 95.1 % — including server processing time ......................................................................................... 0.94 ms  
  — Download 4.2 % ................................................................................................... 0.00 MB/req 8.31 MB/s 0.04 ms  
  ■ Critical  ■ Poor  ■ Ok  ■ Excellent  
```

Gzip/Zstandard enabled, (a bit slower as expected for a benchmark, but still quite good):

```
./vendor/bin/pest stress http://127.0.0.1:8000               

  Stress testing 127.0.0.1:8000 ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― for 5 seconds  
  00:30:47                                                                                                                                    0.93ms  
  00:30:48                                                                                                                                    0.93ms  
  00:30:49                                                                                                                                    0.93ms  
  00:30:50                                                                                                                                    0.93ms  
  00:30:51                                                                                                                                    0.93ms  
  Result ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――  
  Test Duration ............................................................................................................................. 5.04 s  
  Test Concurrency ............................................................................................................................... 1  
  Requests Count ....................................................................................................... 459.31 reqs/s 2315 requests  
  Success Rate ............................................................................................................................. 100.0 %  
  DNS Lookup Duration .....................................................................................................................  0.00 ms  
  TLS Handshake Duration ................................................................................................................... 0.00 ms  
  Request Duration ......................................................................................................................... 0.93 ms  
  — Upload 0.3 % ..................................................................................................... 0.00 MB/req 0.05 MB/s 0.00 ms  
  — TTFB 95.2 % — including server processing time ......................................................................................... 0.89 ms  
  — Download 4.3 % ................................................................................................... 0.00 MB/req 8.42 MB/s 0.04 ms  
  ■ Critical  ■ Poor  ■ Ok  ■ Excellent  
```

```
./vendor/bin/pest stress http://127.0.0.1:8000 -concurrency=8

  Stress testing 127.0.0.1:8000 ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― for 5 seconds  
  00:31:46                                                                                                                                    1.00ms  
  00:31:47                                                                                                                                    0.97ms  
  00:31:48                                                                                                                                    0.96ms  
  00:31:49                                                                                                                                    0.98ms  
  00:31:50                                                                                                                                    0.97ms  
  Result ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――  
  Test Duration ............................................................................................................................. 5.00 s  
  Test Concurrency ............................................................................................................................... 1  
  Requests Count ....................................................................................................... 412.11 reqs/s 2061 requests  
  Success Rate ............................................................................................................................. 100.0 %  
  DNS Lookup Duration .....................................................................................................................  0.00 ms  
  TLS Handshake Duration ................................................................................................................... 0.00 ms  
  Request Duration ......................................................................................................................... 0.97 ms  
  — Upload 0.3 % ..................................................................................................... 0.00 MB/req 0.05 MB/s 0.00 ms  
  — TTFB 95.0 % — including server processing time ......................................................................................... 0.92 ms  
  — Download 4.1 % ................................................................................................... 0.00 MB/req 7.56 MB/s 0.04 ms  
  ■ Critical  ■ Poor  ■ Ok  ■ Excellent  
```

I'll try to make a comparison with Swoole soon.

nunomaduro:
@dunglas Thank you! I will re-run this benchmark again, and let you know how it goes.

nunomaduro:
@dunglas You using `FrankenPHP v1.0.0-rc.3 PHP 8.2.12 Caddy v2.7.5`? I still have the same "slow" results as before...

dunglas:
I'm using the "main" branch with PHP-8.2 dev ZTS compiled locally too. Both compiled locally. I wonder if there is a specific problem with the binary created with static-php-cli that is downloaded by Octane (which is known to be less performant than the Docker images and local builds, but more convenient). I'le dig into that ASAP.

Cluster2a:
> For Symfony, we have benchmarks comparing FrankenPHP to FPM but not to RoadRunner/Swoole as they don't serve the same purpose (the main goal of FrankenPHP is to simplify deployment because there is only 1 service, automatic HTTPS, native HTTP/2 and 3...).

@dunglas, so the main purpose is to have an easier development process? Does it make sense to prioritize a better development experience at the cost of performance?

roberto-butti:
I'm going to take a look as well. Let's see if i can provide some feedback.

ChrisRiddell:
I thought it would be interesting to test the performance by comparing it to herd that's your more typical PHP,Nginx setup and I'm seeing a small performance shortfall on FrankenPHP  but no where close to nunomaduro results but still lower then dunglas results.  Using RC3 on a M1 MBP I wonder if dunglas is on the right track with using a compiled version from the main branch.


Fresh Laravel application, with Breeze/Livewire class API

```php
-- ./vendor/bin/pest stress http://frank.test / https://localhost

  Herd:
  Test Duration ..................................................................................................................... 5.01 s  
  Test Concurrency ....................................................................................................................... 1  
  Requests Count ............................................................................................... 213.61 reqs/s 1070 requests  
  Success Rate ..................................................................................................................... 100.0 %  
  Request Duration ................................................................................................................. 4.82 ms  
  
  Franken PHP:
  Test Duration ..................................................................................................................... 5.00 s  
  Test Concurrency ....................................................................................................................... 1  
  Requests Count ................................................................................................ 148.76 reqs/s 744 requests  
  Success Rate ..................................................................................................................... 100.0 %  
  Request Duration ................................................................................................................. 6.36 ms  
  
-- ./vendor/bin/pest stress http://frank.test / https://localhost --concurrency=8
  
  Herd:
  Test Duration ..................................................................................................................... 5.03 s  
  Test Concurrency ....................................................................................................................... 8  
  Requests Count ................................................................................................ 601.4 reqs/s 3027 requests  
  Success Rate ..................................................................................................................... 100.0 %  
  Request Duration ................................................................................................................ 10.37 ms  

  Franken PHP:
 Test Duration ..................................................................................................................... 5.01 s  
  Test Concurrency ....................................................................................................................... 8  
  Requests Count ............................................................................................... 545.85 reqs/s 2732 requests  
  Success Rate ...................................................................................................................... 99.9 %  
  Request Duration ................................................................................................................ 11.02 ms  
  ```

**Comment by barryvdh** [comment]
I installed a new Laravel project with this fork of Octane, but it does look a bit strange. 

FrankenPHP version:
FrankenPHP v1.0.0-rc.3 PHP 8.2.12 Caddy v2.7.5 h1:HoysvZkLcN2xJExEepaFHK92Qgs7xAiCFydN5x5Hs6Q=%   
Startup:
<details>

```
~/Sites/frankentest/ php artisan octane:start --workers=8

   INFO  Server running…

  Local: http://127.0.0.1:8000 

{"level":"info","ts":1701076094.0864372,"msg":"using provided configuration","config_file":"/Users/barry/Sites/frankentest/Caddyfile","config_adapter":""}
{"level":"warn","ts":1701076094.087071,"msg":"Caddyfile input is not formatted; run 'caddy fmt --overwrite' to fix inconsistencies","adapter":"caddyfile","file":"/Users/barry/Sites/frankentest/Caddyfile","line":29}
{"level":"info","ts":1701076094.087596,"logger":"admin","msg":"admin endpoint started","address":"localhost:2019","enforce_origin":false,"origins":["//localhost:2019","//[::1]:2019","//127.0.0.1:2019"]}
{"level":"info","ts":1701076094.087682,"logger":"tls.cache.maintenance","msg":"started background certificate maintenance","cache":"0x14000144f00"}
{"level":"info","ts":1701076094.112158,"msg":"FrankenPHP started 🐘","php_version":"8.2.12"}
{"level":"info","ts":1701076094.112212,"logger":"tls","msg":"cleaning storage unit","description":"FileStorage:/Users/barry/Library/Application Support/Caddy"}
{"level":"info","ts":1701076094.112244,"logger":"tls","msg":"finished cleaning storage units"}
{"level":"info","ts":1701076094.1122692,"logger":"http.log","msg":"server running","name":"srv0","protocols":["h1","h2","h3"]}
{"level":"info","ts":1701076094.112442,"msg":"autosaved config (load with --resume flag)","file":"/Users/barry/Library/Application Support/Caddy/autosave.json"}
{"level":"info","ts":1701076094.1124449,"msg":"serving initial configuration"}
```
</details>

Laravel Herd
Concurrency 1:   311.40 [#/sec] (mean)
Concurrency 8:  1064.26 [#/sec] (mean)
<details>

```
 ~/Sites/frankentest/ ab -t 5 http://frankentest.test/ 

This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking frankentest.test (be patient)
Finished 1557 requests


Server Software:        nginx/1.22.1
Server Hostname:        frankentest.test
Server Port:            80

Document Path:          /
Document Length:        27515 bytes

Concurrency Level:      1
Time taken for tests:   5.000 seconds
Complete requests:      1557
Failed requests:        0
Total transferred:      44642304 bytes
HTML transferred:       42840855 bytes
Requests per second:    311.40 [#/sec] (mean)
Time per request:       3.211 [ms] (mean)
Time per request:       3.211 [ms] (mean, across all concurrent requests)
Transfer rate:          8719.16 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.1      0       5
Processing:     3    3   1.9      3      40
Waiting:        3    3   1.9      3      40
Total:          3    3   1.9      3      41

Percentage of the requests served within a certain time (ms)
  50%      3
  66%      3
  75%      3
  80%      3
  90%      3
  95%      3
  98%      5
  99%     15
 100%     41 (longest request)
 ~/Sites/frankentest/ ab -t 5 -c 8 http://frankentest.test/ 

This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking frankentest.test (be patient)
Completed 5000 requests
Finished 5322 requests


Server Software:        nginx/1.22.1
Server Hostname:        frankentest.test
Server Port:            80

Document Path:          /
Document Length:        27515 bytes

Concurrency Level:      8
Time taken for tests:   5.001 seconds
Complete requests:      5322
Failed requests:        0
Total transferred:      152621056 bytes
HTML transferred:       146462345 bytes
Requests per second:    1064.26 [#/sec] (mean)
Time per request:       7.517 [ms] (mean)
Time per request:       0.940 [ms] (mean, across all concurrent requests)
Transfer rate:          29804.90 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       1
Processing:     3    7   6.1      6      82
Waiting:        3    7   6.1      6      82
Total:          3    7   6.1      6      82

Percentage of the requests served within a certain time (ms)
  50%      6
  66%      7
  75%      7
  80%      8
  90%      9
  95%     11
  98%     28
  99%     44
 100%     82 (longest request)
```
</details>

Octane (without workers)
Concurrency 1:   197.13 [#/sec] (mean)
Concurrency 8:   207.76 [#/sec] (mean)

<details>

```
 ~/Sites/frankentest/ ab -t 5 -c 1 http://127.0.0.1:8000/ 

This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient)
Finished 990 requests


Server Software:        Caddy
Server Hostname:        127.0.0.1
Server Port:            8000

Document Path:          /
Document Length:        27515 bytes

Concurrency Level:      1
Time taken for tests:   5.022 seconds
Complete requests:      990
Failed requests:        0
Total transferred:      28336770 bytes
HTML transferred:       27239850 bytes
Requests per second:    197.13 [#/sec] (mean)
Time per request:       5.073 [ms] (mean)
Time per request:       5.073 [ms] (mean, across all concurrent requests)
Transfer rate:          5510.16 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:     1    5  30.9      1     392
Waiting:        1    2   8.0      1      78
Total:          1    5  30.9      1     392

Percentage of the requests served within a certain time (ms)
  50%      1
  66%      1
  75%      1
  80%      1
  90%      2
  95%      3
  98%     57
  99%    166
 100%    392 (longest request)
 
  ~/Sites/frankentest/ ab -t 5 -c 8 http://127.0.0.1:8000/   

This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient)
Finished 1099 requests


Server Software:        Caddy
Server Hostname:        127.0.0.1
Server Port:            8000

Document Path:          /
Document Length:        27515 bytes

Concurrency Level:      8
Time taken for tests:   5.290 seconds
Complete requests:      1099
Failed requests:        0
Total transferred:      31657038 bytes
HTML transferred:       30431590 bytes
Requests per second:    207.76 [#/sec] (mean)
Time per request:       38.506 [ms] (mean)
Time per request:       4.813 [ms] (mean, across all concurrent requests)
Transfer rate:          5844.38 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.2      0       1
Processing:     1   35 120.8      2     520
Waiting:        1    4  11.9      2     127
Total:          1   35 120.8      2     520

Percentage of the requests served within a certain time (ms)
  50%      2
  66%      2
  75%      3
  80%      4
  90%      9
  95%    480
  98%    493
  99%    495
 100%    520 (longest request)
```
</details>

Octane (8 workers)
Concurrency 1:   190.05 [#/sec] (mean)
Concurrency 8:   208.91 [#/sec] (mean)

<details>

 ~/Sites/frankentest/ ab -t 5 -c 1 http://127.0.0.1:8000/ 
```
This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient)
Finished 993 requests


Server Software:        Caddy
Server Hostname:        127.0.0.1
Server Port:            8000

Document Path:          /
Document Length:        27515 bytes

Concurrency Level:      1
Time taken for tests:   5.225 seconds
Complete requests:      993
Failed requests:        0
Total transferred:      28422639 bytes
HTML transferred:       27322395 bytes
Requests per second:    190.05 [#/sec] (mean)
Time per request:       5.262 [ms] (mean)
Time per request:       5.262 [ms] (mean, across all concurrent requests)
Transfer rate:          5312.18 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:     1    5  28.2      1     346
Waiting:        1    3  13.5      1     263
Total:          1    5  28.2      1     346

Percentage of the requests served within a certain time (ms)
  50%      1
  66%      1
  75%      1
  80%      1
  90%      2
  95%      4
  98%     67
  99%    150
 100%    346 (longest request)
 ~/Sites/frankentest/ ab -t 5 -c 8 http://127.0.0.1:8000/ 

This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient)
Finished 1101 requests


Server Software:        Caddy
Server Hostname:        127.0.0.1
Server Port:            8000

Document Path:          /
Document Length:        27515 bytes

Concurrency Level:      8
Time taken for tests:   5.270 seconds
Complete requests:      1101
Failed requests:        0
Total transferred:      31542546 bytes
HTML transferred:       30321530 bytes
Requests per second:    208.91 [#/sec] (mean)
Time per request:       38.295 [ms] (mean)
Time per request:       4.787 [ms] (mean, across all concurrent requests)
Transfer rate:          5844.67 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.2      0       1
Processing:     1   38 124.6      2     519
Waiting:        1    5  15.3      2     128
Total:          1   38 124.6      2     520

Percentage of the requests served within a certain time (ms)
  50%      2
  66%      3
  75%      3
  80%      4
  90%     10
  95%    482
  98%    486
  99%    495
 100%    520 (longest request)
```
</details>

When I run the same binary standalone with the Cadyfile from https://github.com/dunglas/frankenphp/blob/main/docs/laravel.md

```
{
	frankenphp
	order php_server before file_server
}

# The domain name of your server
http://127.0.0.1:8001 {
    # Enable compression (optional)
    encode zstd gzip
    # Execute PHP files in the current directory and serve assets
    php_server
}
```

I get the same results with concurrency 1, but faster results with concurrency 8

Octane standalone, default config
Concurrency 1:   189.70 [#/sec] (mean)
Concurrency 8:   878.19 [#/sec] (mean)
<details>

```
 ~/Sites/frankentest/ ab -t 5 -c 1 http://127.0.0.1:8001/ 

This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient)
Finished 949 requests


Server Software:        Caddy
Server Hostname:        127.0.0.1
Server Port:            8001

Document Path:          /
Document Length:        27515 bytes

Concurrency Level:      1
Time taken for tests:   5.003 seconds
Complete requests:      949
Failed requests:        0
Total transferred:      27163227 bytes
HTML transferred:       26111735 bytes
Requests per second:    189.70 [#/sec] (mean)
Time per request:       5.272 [ms] (mean)
Time per request:       5.272 [ms] (mean, across all concurrent requests)
Transfer rate:          5302.50 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:     3    5  14.2      3     299
Waiting:        3    5  14.2      3     299
Total:          3    5  14.2      3     299

Percentage of the requests served within a certain time (ms)
  50%      3
  66%      3
  75%      4
  80%      4
  90%      4
  95%      4
  98%     18
  99%     81
 100%    299 (longest request)
 ~/Sites/frankentest/ ab -t 5 -c 8 http://127.0.0.1:8001/ 

This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient)
Finished 4392 requests


Server Software:        Caddy
Server Hostname:        127.0.0.1
Server Port:            8001

Document Path:          /
Document Length:        27515 bytes

Concurrency Level:      8
Time taken for tests:   5.001 seconds
Complete requests:      4392
Failed requests:        7
   (Connect: 0, Receive: 0, Length: 7, Exceptions: 0)
Non-2xx responses:      7
Total transferred:      125559665 bytes
HTML transferred:       120699538 bytes
Requests per second:    878.19 [#/sec] (mean)
Time per request:       9.110 [ms] (mean)
Time per request:       1.139 [ms] (mean, across all concurrent requests)
Transfer rate:          24517.48 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.1      0       2
Processing:     4    9  24.3      5     275
Waiting:        3    9  24.2      5     275
Total:          4    9  24.3      6     275

Percentage of the requests served within a certain time (ms)
  50%      6
  66%      6
  75%      7
  80%      7
  90%      8
  95%      9
  98%     12
  99%    196
 100%    275 (longest request)
```
</details>

I'm not sure how to start the worker mode without Laravel, because I need a script for that. The existing script gives an error:
> 2023/11/27 09:27:57.926	ERROR	unexpected termination, restarting	{"worker": "/Users/barry/Sites/frankentest/public/frankenphp-worker.php"}

**Comment by barryvdh** [comment]
Ah I got the stand-alone script working using an .env file to set the Base path, but got similar results using 0 or 8 workers.

But it seems to be a bit weird that the non-worker mode standalone is faster then Octane, with or without workers, or even standalone with worker script..

(And that Laravel Herd is the fastest in all tests, especially with concurrent requests)

```
frankenphp {
        worker frankenphp-worker.php 8
    }
```

FrankenPHP, 9 workers:
Concurrency 1: 159.14 [#/sec] (mean)
Concurrency 8:   137.80 [#/sec] (mean)

<details>

```
~/Sites/frankentest/ ab -t 5 -c 1 http://127.0.0.1:8001/ 

This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient)
Finished 796 requests


Server Software:        Caddy
Server Hostname:        127.0.0.1
Server Port:            8001

Document Path:          /
Document Length:        27515 bytes

Concurrency Level:      1
Time taken for tests:   5.002 seconds
Complete requests:      796
Failed requests:        0
Total transferred:      22783908 bytes
HTML transferred:       21901940 bytes
Requests per second:    159.14 [#/sec] (mean)
Time per request:       6.284 [ms] (mean)
Time per request:       6.284 [ms] (mean, across all concurrent requests)
Transfer rate:          4448.22 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:     3    6  22.2      3     193
Waiting:        3    6  22.2      3     193
Total:          3    6  22.2      3     193

Percentage of the requests served within a certain time (ms)
  50%      3
  66%      3
  75%      3
  80%      3
  90%      4
  95%      5
  98%     14
  99%    183
 100%    193 (longest request)
~/Sites/frankentest/ ab -t 5 -c 8 http://127.0.0.1:8001/ 

This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient)
Finished 689 requests


Server Software:        Caddy
Server Hostname:        127.0.0.1
Server Port:            8001

Document Path:          /
Document Length:        27515 bytes

Concurrency Level:      8
Time taken for tests:   5.000 seconds
Complete requests:      689
Failed requests:        0
Total transferred:      19721247 bytes
HTML transferred:       18957835 bytes
Requests per second:    137.80 [#/sec] (mean)
Time per request:       58.057 [ms] (mean)
Time per request:       7.257 [ms] (mean, across all concurrent requests)
Transfer rate:          3851.69 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       1
Processing:    19   58  71.6     26     396
Waiting:       19   58  71.6     26     396
Total:         20   58  71.6     27     396

Percentage of the requests served within a certain time (ms)
  50%     27
  66%     27
  75%     27
  80%     28
  90%    215
  95%    220
  98%    222
  99%    226
 100%    396 (longest request)


```
</details>

FrankenPHP, 8 workers:
Concurrency 1: 150.99 [#/sec] (mean)
Concurrency 8:  322.78 [#/sec] (mean)

<details>

```
 ~/Sites/frankentest/ ab -t 5 -c 1 http://127.0.0.1:8001/ 

This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient)
Finished 768 requests


Server Software:        Caddy
Server Hostname:        127.0.0.1
Server Port:            8001

Document Path:          /
Document Length:        27515 bytes

Concurrency Level:      1
Time taken for tests:   5.086 seconds
Complete requests:      768
Failed requests:        0
Total transferred:      21982464 bytes
HTML transferred:       21131520 bytes
Requests per second:    150.99 [#/sec] (mean)
Time per request:       6.623 [ms] (mean)
Time per request:       6.623 [ms] (mean, across all concurrent requests)
Transfer rate:          4220.50 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       1
Processing:     3    6  22.2      3     173
Waiting:        3    6  22.2      3     173
Total:          3    7  22.2      3     173

Percentage of the requests served within a certain time (ms)
  50%      3
  66%      4
  75%      4
  80%      4
  90%      4
  95%      5
  98%     18
  99%    169
 100%    173 (longest request)
 ~/Sites/frankentest/ ab -t 5 -c 8 http://127.0.0.1:8001/ 

This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient)
Finished 1614 requests


Server Software:        Caddy
Server Hostname:        127.0.0.1
Server Port:            8001

Document Path:          /
Document Length:        27515 bytes

Concurrency Level:      8
Time taken for tests:   5.000 seconds
Complete requests:      1614
Failed requests:        0
Total transferred:      46197522 bytes
HTML transferred:       44409210 bytes
Requests per second:    322.78 [#/sec] (mean)
Time per request:       24.785 [ms] (mean)
Time per request:       3.098 [ms] (mean, across all concurrent requests)
Transfer rate:          9022.27 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.1      0       1
Processing:     7   24  31.4     15     266
Waiting:        7   24  31.4     15     266
Total:          8   25  31.4     15     266

Percentage of the requests served within a certain time (ms)
  50%     15
  66%     24
  75%     25
  80%     25
  90%     26
  95%     40
  98%    173
  99%    226
 100%    266 (longest request)
```

</details>

For comparison, Roadrunner is slightly faster:

Octane Roadrunner
Concurrency 1:  264.08 [#/sec] (mean)
Concurrency 8:  377.76 [#/sec] (mean)

<details>

```details
/Sites/frankentest/public/ ab -t 5 -c 1 http://127.0.0.1:8000/ 

This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient)
Finished 1333 requests


Server Software:        
Server Hostname:        127.0.0.1
Server Port:            8000

Document Path:          /
Document Length:        27515 bytes

Concurrency Level:      1
Time taken for tests:   5.048 seconds
Complete requests:      1333
Failed requests:        0
Total transferred:      38049152 bytes
HTML transferred:       36677495 bytes
Requests per second:    264.08 [#/sec] (mean)
Time per request:       3.787 [ms] (mean)
Time per request:       3.787 [ms] (mean, across all concurrent requests)
Transfer rate:          7361.09 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       1
Processing:     1    4  13.8      1     150
Waiting:        1    4  13.6      1     150
Total:          1    4  13.8      1     150

Percentage of the requests served within a certain time (ms)
  50%      1
  66%      1
  75%      1
  80%      1
  90%      2
  95%      2
  98%     76
  99%     81
 100%    150 (longest request)
 ~/Sites/frankentest/public/ ab -t 5 -c 8 http://127.0.0.1:8000/ 

This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient)
Finished 2005 requests


Server Software:        
Server Hostname:        127.0.0.1
Server Port:            8000

Document Path:          /
Document Length:        27515 bytes

Concurrency Level:      8
Time taken for tests:   5.308 seconds
Complete requests:      2005
Failed requests:        0
Total transferred:      57247104 bytes
HTML transferred:       55182930 bytes
Requests per second:    377.76 [#/sec] (mean)
Time per request:       21.178 [ms] (mean)
Time per request:       2.647 [ms] (mean, across all concurrent requests)
Transfer rate:          10532.98 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.2      0       7
Processing:     1   20  85.5      2     543
Waiting:        1   19  83.4      2     542
Total:          1   20  85.5      2     543

Percentage of the requests served within a certain time (ms)
  50%      2
  66%      3
  75%      3
  80%      3
  90%      5
  95%     10
  98%    456
  99%    478
 100%    543 (longest request)
```

</details>

src/Commands/StartFrankenPhpCommand.php
```diff
@@ -0,0 +1,215 @@
...
+            ->filter()
+            ->each(function ($output) {
+                $this->raw($output);
+            });
+
+        Str::of($errorOutput)
+            ->explode("\n")
+            ->filter()
+            ->each(function ($output) {
+                $this->raw($output);
```
barryvdh, line 165: This doesn't work properly with --quiet. 
To format it nicely and show requests better, we could do something like this:

```php
        Str::of($errorOutput)
            ->explode("\n")
            ->filter()
            ->each(function ($output) {

                if (Str::contains($output, "http.log.access") && Str::contains($output, "handled request") ) {
                    if ($debug = json_decode(trim(substr($output, strpos($output, "handled request") + 15)), true)) {

                        if (isset($debug['duration'])
                            && isset($debug['request']['method'])
                            && isset($debug['request']['uri'])
                            && isset($debug['status'])) {

                            return $this->requestInfo([
                                'method' => $debug['request']['method'],
                                'url' => $debug['request']['uri'],
                                'statusCode' => $debug['status'],
                                'duration' => $debug['duration'],
                            ]);
                        }

                    }
                }

                $debug = json_decode($output, true);
                if (is_array($debug) && isset($debug['level']) && isset($debug['msg'])) {
                    $extra = Arr::except($debug, ['logger', 'level', 'msg', 'ts']);
                    $msg = (isset($debug['logger']) ? '[' .$debug['logger'] .'] ' : '') .$debug['msg'] . ' ' . ($extra ? json_encode($extra) : '');
                    switch ($debug['level']) {
                        case 'error':
                            return $this->error($msg);
                        case 'warn':
                            return $this->warn($msg);
                        default:
                            return $this->info($msg);
                    }
                }

                $this->raw($output);
            });
```

image

barryvdh:
haha bummer you beat me to it, was about to post that as well :P 

Kept removing things until it matched better. With concurrency 8 went from 200 to 500+ req/s

<details>
```
~/Sites/frankentest/ ab -t 5 -c 1 http://127.0.0.1:8000/ 

This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient)
Finished 1068 requests


Server Software:        Caddy
Server Hostname:        127.0.0.1
Server Port:            8000

Document Path:          /
Document Length:        27515 bytes

Concurrency Level:      1
Time taken for tests:   5.100 seconds
Complete requests:      1068
Failed requests:        0
Total transferred:      30569364 bytes
HTML transferred:       29386020 bytes
Requests per second:    209.41 [#/sec] (mean)
Time per request:       4.775 [ms] (mean)
Time per request:       4.775 [ms] (mean, across all concurrent requests)
Transfer rate:          5853.58 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:     1    5  29.5      1     244
Waiting:        1    5  29.5      1     244
Total:          1    5  29.5      1     244

Percentage of the requests served within a certain time (ms)
  50%      1
  66%      1
  75%      1
  80%      1
  90%      1
  95%      1
  98%      5
  99%    228
 100%    244 (longest request)
 ~/Sites/frankentest/ ab -t 5 -c 8 http://127.0.0.1:8000/ 

This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient)
Finished 2340 requests


Server Software:        Caddy
Server Hostname:        127.0.0.1
Server Port:            8000

Document Path:          /
Document Length:        27515 bytes

Concurrency Level:      8
Time taken for tests:   5.001 seconds
Complete requests:      2340
Failed requests:        0
Total transferred:      66977820 bytes
HTML transferred:       64385100 bytes
Requests per second:    467.95 [#/sec] (mean)
Time per request:       17.096 [ms] (mean)
Time per request:       2.137 [ms] (mean, across all concurrent requests)
Transfer rate:          13080.16 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.1      0       2
Processing:     1   16  94.3      2     700
Waiting:        1   16  94.3      1     700
Total:          1   16  94.3      2     700

Percentage of the requests served within a certain time (ms)
  50%      2
  66%      2
  75%      2
  80%      2
  90%      3
  95%      4
  98%    625
  99%    645
 100%    700 (longest request)

```
</details>

barryvdh:
So that looks better, but not really sure why the worker mode is just as fast as the non-worker (or stand-alone index.php). I would expect that the biggest gains for Octane would come from that?

barryvdh:
It looks like it keeps restarting the worker script each request, which doesn't seem to be the intention. I've added some logs and `$nbRequest` keeps on 0, and the `$worker` keeps being null every time, so a fresh application is started.
Will keep looking to see why it's terminated.
I don't see any logs calling the terminate function, so it would seem the application dies on itself, never getting to the next iteration of the loop.

barryvdh:
Okay lunch break is over, but if I use this simple worker script:

```php
<?php

if ((! ($_SERVER['FRANKENPHP_WORKER'] ?? false)) || ! function_exists('frankenphp_handle_request')) {
    echo 'You need FrankenPHP in worker mode to use this script.';

    exit(1);
}

ignore_user_abort(true);

$id = uniqid();
$nbRequests = 0;
$handler = static function () use ($nbRequests, $id) {
    echo  'NB Requests: ' .$nbRequests . ', id: ' .$id;
};

while (\frankenphp_handle_request($handler)) {
    $nbRequests++;
}
```

That gives me 0 requests and a new id everytime. That would indicate that every time the worker is called, a fresh instance is used, instead of re-using that one. But as far is I can tell, that doesn't really have anything to do with Octane itself, more with FrankenPHP? Or maybe something Macbook specific? (I have an M2 Pro)

dunglas:
@barryvdh thanks for the numbers.

By default, the number of workers is 2 times the numbers of CPUs of the machine (so you likely have 10 CPUs).
Indeed you need to send several requests to "warm" all the workers to get the performance benefit of the worker mode.

dunglas:
I've found the performance issue. It's because the output of Caddy was piped to Symfony Console, and Symfony Console adds a huge overhead (I don't know exactly why). https://github.com/laravel/octane/pull/764/commits/e36cac6a047461c4f639f3a14bd53fbf0785b269 fixes the issue.

Before the patch (on my Mac): `4.01 reqs/s`
After: `616.75 resq/s`

Performance can be improved by disabling logs entirely or logging in a file (config in the Caddyfile), but logging on stderr is convenient in dev and for 12-factor apps.

@nunomaduro, I think that this is ready for another round of reviews!

taylorotwell:
Thanks for working on this @dunglas 👍

barryvdh:
> I've found the performance issue. It's because the output of Caddy was piped to Symfony Console, and Symfony Console adds a huge overhead (I don't know exactly why). [e36cac6](https://github.com/laravel/octane/commit/e36cac6a047461c4f639f3a14bd53fbf0785b269) fixes the issue.

@dunglas why would interactive mode need to be used at all? Can it not be always non-interactive?

> Performance can be improved by disabling logs entirely or logging in a file (config in the Caddyfile), but logging on stderr is convenient in dev and for 12-factor apps.

You could use the verbosity levels similar to the other servers. Log the output as json instead of formatted as a filter, and parse it in PHP like Roadrunner/Swoole?

```
	log {
		output stdout
		format json
	}
```

```php
    protected function writeServerOutput($server)
    {
        [$output, $errorOutput] = $this->getServerOutput($server);

        Str::of($output . "\n" . $errorOutput)
            ->explode("\n")
            ->filter()
            ->each(function ($output) {
                if (! is_array($debug = json_decode($output, true))) {
                    return $this->info($output);
                }

                if (!isset($debug['level'])) {
                    return $this->raw($output);
                }

                if ($debug['level'] == 'info'
                    && isset($debug['request']['uri'])
                    && $debug['logger'] == 'http.log.access.log0') {

                    return $this->requestInfo([
                        'method' => $debug['request']['method'],
                        'url' => $debug['request']['uri'],
                        'statusCode' => $debug['status'] ?? 0,
                        'duration' => $debug['duration'] ?? 0,
                    ]);
                }

                if ($debug['level'] === 'info') {
                    return $this->info($output);
                } elseif ($debug['level'] === 'warn') {
                    return $this->warn($output);
                } else {
                    return $this->error($output);
                }
            });
    }
```

Then running the server with `--quiet` would prevent the output of all the requests, similar to the others.

flexchar:
I just heard about FrankenPHP from https://laravel-news.com/frankenphp. I wanted to come and say big thank you for developing this for PHP & Laravel community as a whole. Seeing all the attempts and crazy developments in the world of JavaScript, it's so impressive and proud to know that we're about to have 3 solutions for running laravel as fast as JS on the edge. Wow. Seriously, thank you Kévin Dunglas and the laravel team for working on getting this on. Definitely.

src/Commands/StartCommand.php
```diff
@@ -23,6 +23,8 @@ class StartCommand extends Command implements SignalableCommandInterface
                     {--task-workers=auto : The number of task workers that should be available to handle tasks}
                     {--max-requests=500 : The number of requests to process before reloading the server}
                     {--rr-config= : The path to the RoadRunner .rr.yaml file}
+                    {--frankenphp-config= : The path to the FrankenPHP Caddyfile file}
+                    {--https : Enable HTTPS, HTTP/2 and HTTP/3, automatically generate and renew certificates}
```
rodrigopedra, line 27: Maybe it is worth adding a note that the `--https` option only applies to FrakenPHP?

dunglas:
@barryvdh good idea to disable logs when the quiet option is set. I'll update the PR. Regarding the "tty" mode, it allows Caddy to pass in "user-friendly" mode and to display human-readable logs. I find this feature helpful when developing and there is no way to "re-code" it in Octane as it's an upstream feature. In prod, JSON is logged "as usual".

roberto-butti:
amazing performance improvement with the latest changes in the past 3 weeks.
Please try your tests with Octane::route 🚀

```
use Laravel\Octane\Facades\Octane;
use Symfony\Component\HttpFoundation\Response;

Octane::route('GET', '/route-franken', function () {
    return new Response('Hi Folks, we have "Octane Route".');
});
Route::get(
'/route-standard', function () {
    return new Response('Hi Folks, we have "Standard Route".');
});
```
I am testing some functionality with multiple workers. 🎉

taylorotwell:
@dunglas I've reviewed the entire PR and it's pretty much ready to go. 

I'm a bit confused on the recent discussions around output. To clarify, how should Octane + FrankenPHP actually be run in production so that the output doesn't bog down the application? 

@nunomaduro should we try to add some sort of pretty printing treatment to FrankenPHP like we do RoadRunner / Swoole for local development?

dunglas:
Thanks for the review @taylorotwell!

Caddy (and so FrankenPHP) has a powerful (and fast) native logger that has a lot of options including pretty printing in dev mode, sampling and many backends:

* https://caddyserver.com/docs/logging
* https://caddyserver.com/docs/caddyfile/directives/log

In my opinion, delegating logging entirely to Caddy would be the best option, but it's not how the RoadRunner/Swoole integrations work. What do you think?

francislavoie:
I don't think the `default` (global) logger should default to `ERROR` in prod, it should still default to `INFO`, otherwise important startup messages and deprecation warnings will be lost.

I think access logs (`log` directive) could be off by default in production though (or access log's level at `ERROR` potentially), because that can be very noisy on a high-traffic site and it might not be necessary for everyone to have enabled.

taylorotwell:
@dunglas you have most experience with FrankenPHP both in dev and production, so I think we should defer to you on the best setup here. Is this PR currently setup in the way you think it should optimized?

I noticed when i ran it locally I was seeing basically access log style output in my console - is that the output you mentioned was bottlenecking things? Or is that bottleneck totally removed? Would I run `octane:start` differently in production to maximize speed?

dunglas:
@francislavoie reverted

@taylorotwell the bottleneck is now totally removed if the terminal is a TTY (and the performance issue is less likely to happen if it isn't, but I need to double-check this).

I just pushed a commit that simplifies the logging-related code and should improve performance when the terminal isn't a TTY. I would like to avoid using `$server->getIncrementalErrorOutput()` entirely (even if not a TTY) and pipe Caddy directly to stderr, but I'm not sure if it's possible with Symfony Console. I'm looking into this.

dunglas:
This doesn't look possible to write directly to stdout if not in TTY mode with Symfony Process, so it's the best we can do without patching Symfony.

src/Commands/StartFrankenPhpCommand.php
```diff
@@ -0,0 +1,231 @@
...
+                        'request' => $request,
+                    ] = $debug;
+
+                    return $this->requestInfo([
+                        'method' => $method,
+                        'url' => $url,
+                        'statusCode' => $statusCode,
+                        'duration' => (float) $duration * 1000,
+                    ]);
+                }
```
barryvdh, line 216: This is missing the regular output (the startup/error logs are also json and are now ignored)

nunomaduro: Yeah - working on it.

nunomaduro:
Hi @dunglas, I am currently reviewing this pull request, as you may know, and here is the list of issues that need to be addressed before we can merge it:

1. Sometimes, when specifying a random number of workers, the server crashes: `php octane:start —workers=16`.
2. Reloading the server causes it to crash: `php artisan octane:reload && php artisan octane:reload && php artisan octane:reload && php artisan octane:reload && php artisan octane:reload && php artisan octane:reload && php artisan octane:reload && php artisan octane:reload && php artisan octane:reload`.
3. Occasionally, the server does not start after multiple attempts at stopping and starting. This can be reproduced by repeatedly pressing Ctrl+C followed by `php artisan octane:start`.
4. Do the caddyfile and the worker file really need to be at the application level? If so, is there any reason why they are not part of version control (git)?

josecl:
Great work @dunglas !

It is possible include support for `pgsql` by default in the `frankenphp` binary?

Another thing...
I was not able to run the tests using the `frankenphp` binary. I got these errors on my M1:

```bash
./frankenphp php-cli vendor/bin/pest

Fatal error: Namespace declaration statement has to be the very first statement or after any declare call in the script in /Users/.../vendor/bin/pest on line 13
```

```bash
./frankenphp php-cli artisan test

#!/usr/bin/env php
sh: line 0: exec: : not found
```

It would be great to test Laravel with the same binary used on deployed.

dunglas:
@josecl thanks!

Postgres support is available on Linux and in Docker images, but it is currently disabled on Mac because of this issue: https://github.com/crazywhalecc/static-php-cli/issues/278

For the scripts, indeed FrankenPHP currently don't handle shebangs, I'll work on this!

nunomaduro (approved):

oprypkhantc:
Hey @dunglas. We're currently running Octane/Swoole + Nginx. If we were to switch Nginx with Caddy (so Swoole + Caddy), how would that be different from FrankenPHP? The proxy configuration is already extremely simple in Caddy, and the Caddy process itself is lightweight, so what are the practical benefits of this complex integration?

dunglas:
Hi @oprypkhantc, the main difference is that with FrankenPHP, Caddy isn't acting as a proxy: FrankenPHP is a custom build of Caddy with an extra module that embeds PHP in it. That means that PHP and Caddy are the same program, running as the same process. Calling the PHP engine from Caddy is just a function call, with no network connection involved, no IPC, etc.

It's probably explained more clearly at the beginning of this video: https://www.youtube.com/watch?v=q6FQaaFZVy4

oprypkhantc:
> Hi @oprypkhantc, the main difference is that with FrankenPHP, Caddy isn't acting as a proxy: FrankenPHP is a custom build of Caddy with an extra module that embeds PHP in it. That means that PHP and Caddy are the same program, running as the same process. Calling the PHP engine from Caddy is just a function call, with no network connection involved, no IPC, etc.
> 
> 
> 
> It's probably explained more clearly at the beginning of this video: https://www.youtube.com/watch?v=q6FQaaFZVy4

Thank you. It'd be interesting to see if this affects performance for our API-only app though, as there's not a lot of traffic, so likely not a lot of overhead from the proxying.

I'll post the results when I get to try it if anyone is interested.

francislavoie:
@fannyfan414 see https://caddyserver.com/docs/automatic-https#storage, you're meant to configure storage to be shared across all your instances. You can use filesystem syncing, or you can use a Redis storage module, for example.

nickarellano:
Very interested in this. How does this effect running it as a binary with octane?

jhoanborges:
Waiting for this...

ARehmanMahi:
@dunglas Would be really nice to see a section in docs for included php extensions. 

Plus wondering how to know which php versiond would be supported by franky. how would updates be handeled, whwre can I find info on this?

marsderp:
> Very interested in this. How does this effect running it as a binary with octane?

Second this - anyone know of any guides on how to build a binary with octane yet?

bsienn-khan:
Sorry for the noob question, when using `php artisan octane:install --server=frankenphp`, what's the difference between using system php rather than frankenphp's php-cli like:
  
  `php artisan octane:start VS ./frankenphp php-cli artisan octane:start`

dunglas:
@ARehmanMahi this will use the PHP engine embedded in the binary. This should work, even with no local installation of PHP.

bsienn-khan:
> @ARehmanMahi this will use the PHP engine embedded in the binary. This should work, even with no local installation of PHP.

@dunglas Thanks for the reply, one more question as I'm a bit unclear.

As FrankenPHP has its own PHP, I assume `php artisan octane:start` will use Franky's built-in PHP to run the app **even though it was started with system's php**, is this how it works?

P.S. If that's the case, I can completely get rid of systems php for the project I reckon. Which feels great.

rodrigopedra:
@ARehmanMahi if you alias `php` to the FrankenPHP binary in your system, that will work as you described.

Otherwise, your system won't know which binary to run when you type `php artisan ...`.

Also, keep in mind that composer will try to resolve the PHP binary, so you still need a globally accessible PHP binary, be it an alias to the FrankenPHP binary or a system installed one.

whousehomestay:
Is this still in beta version like in the [Octane Docs](https://laravel.com/docs/10.x/octane#frankenphp) until now?

jameshulse:
> Is this still in beta version like in the [Octane Docs](https://laravel.com/docs/10.x/octane#frankenphp) until now?

I'm wondering the same, and whether there is somewhere that we can track progress of this integration moving out of beta. We are holding off moving to octane until this integration stabilises as it seems to be the most promising in a number of ways.

oprypkhantc:
> > Is this still in beta version like in the [Octane Docs](https://laravel.com/docs/10.x/octane#frankenphp) until now?
> 
> I'm wondering the same, and whether there is somewhere that we can track progress of this integration moving out of beta. We are holding off moving to octane until this integration stabilises as it seems to be the most promising in a number of ways.

Speaking from experience, you'll likely have to fix countless little bugs when moving to Octane, regardless of what specific web server you'll be using, be it RoadRunner, Swoole or FrankenPHP. Octane itself is stable as far as I know and as far as I've seen in the last year we've been running it in production.

Switching the integration from Swoole to FrankenPHP should not touch any of your application code and should require only minimal changes to your (hopefully containerized) setup, so consider moving to Octane/Swoole first and only then to FrankenPHP :)

bsienn-khan:
@oprypkhantc your 1st & 2nd line seems contradictory. 

Is it moving to Octane that will cause countless bugs? 
Or not moving and starting with Octane from scratch is the better option. 
I can't deduce what you mean actually.

Moving to in-memory application architecture requires a keen eye on using STATIC/Singleton, which could cause bugs or memory leaks. Note to self.

oprypkhantc:
@ARehmanMahi Moving to Octane itself is likely to cause bugs. 

And the problem is not just static/singletons; we had more problems with Laravel itself than with our app code. There is [a lot that Octane does to "fix"](https://github.com/laravel/octane/tree/2.x/src/Listeners) the poor implementation choices of Laravel; for example, the container/app instance is cloned for every request, meaning you have to be extra careful when injecting the `Container` in your services. Same goes for multiple other things too, to which Octane has a whole dedicated [section in the docs](https://laravel.com/docs/10.x/octane#dependency-injection-and-octane).

bsienn-khan:
Okay, so you mean it's not necessarily the app or octane, but laravel itself. 
Well, I can agree on this with you. Recently been hit by weird issues in laravel. Cache maco plus the tags memory leaks for instance. Good luck everyone.

PR #775, opened 2023-12-19
[2.x] add missing host for http option

Jubeki:


nunomaduro:
@Jubeki can you elaborate on why you think this fix is necessary?

Jubeki:
Okay, my wording was a bad choice. I am not sure if it is a fix, but for me it is missing the host option, and I would assume that if you specify another host for some reason, it would be missing with the http option.

If you already have the host available, why not use it there too.

nunomaduro:
I see. In fact, we made this change because typically, in an non-http scenarios, you don't want Caddy to validate the host. E.g: using NGINX proxy features, or even locally using a different domain.

PR #777, opened 2023-12-20
[2.x] Adds support for multiple FrankenPHP Octane instances

nunomaduro:
This pull request adds support for multiple FrankenPHP Octane instances.


src/Commands/StartFrankenPhpCommand.php
```diff
@@ -149,6 +179,18 @@ protected function buildMercureConfig()
...
     }
 
+    /**
+     * Get the port the admin URL should be available on.
+     *
+     * @return int
+     */
+    protected function adminPort()
+    {
+        $defaultPort = 2019;
```
dunglas, line 189: Could be a class constant?

dunglas (commented):
LGTM

src/Commands/StartFrankenPhpCommand.php
nunomaduro: Yeah - keeping it as variable for now.

Issue #774, opened 2023-12-19
Unable to determine the current FrankenPHP binary version \ unable to get instance ID; storage clean stamps will be incomplete


HDVinnie:
### Octane Version

^2.2

### Laravel Version

10.37.3

### PHP Version

8.3.0

### What server type are you using?

FrankenPHP

### Server Version

Ubuntu 22.04.1

### Database Driver & Version

MySQL 8.0.35

### Description

php artisan octane:install

 Which application server you would like to use?:
  [0] roadrunner
  [1] swoole
  [2] frankenphp
 > 2

 FrankenPHP's Octane integration is in beta and should be used with caution in production. Do you wish to continue? (yes/no) [no]:
 > yes

 Unable to locate FrankenPHP binary. Should Octane download the binary for your operating system? (yes/no) [yes]:
 > yes

 116990064/116990064 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

   INFO  Octane installed successfully.

Then upon starting octane.

php artisan octane:start

   WARN  Unable to determine the current FrankenPHP binary version. Please report this issue: https://github.com/laravel/octane/issues/new.

   INFO  Server running…

  Local: http://127.0.0.1:8000 

  Press Ctrl+C to stop the server

   WARN  unable to get instance ID; storage clean stamps will be incomplete

### Steps To Reproduce

Install octane, select frakenphp, configure config and nginx, start octane.

HDVinnie:
I should note that the site is accessible and does seem to be working fine regardless of the two WARN messages.

cbpatel17:
And it's also crashing without any information

phh:
Did the same thing today in Laravel Sail.

Also how to reproduce on a fresh install:

```
curl -s https://laravel.build/octane-frankenphp | bash
cd octane-frankenphp && ./vendor/bin/sail up -d
./vendor/bin/sail composer require laravel/octane
./vendor/bin/sail artisan octane:install --server=frankenphp

yes
yes

./vendor/bin/sail artisan octane:start
```

binaryfire:
Hi @HDVinnie. Did this solve the `unable to get instance ID; storage clean stamps will be incomplete` part of the issue for you? I'm still seeing that on startup (as are others: https://github.com/laravel/octane/issues/810).

HDVinnie:
[...]

Issue #771, opened 2023-12-19
Your FrankenPHP binary version may be incompatible with Octane.


riasvdv:
### Octane Version

2.2.0

### Laravel Version

10.37.3

### PHP Version

8.2.12

### What server type are you using?

FrankenPHP

### Server Version

1.0.1

### Description

1. The issue template didn't let me select Frankenphp as the server type.

2. When letting Octane download the Frankenphp binary on an Ubuntu server, the returned version by Frankenphp (`b2e92a673a8a5e1c285b6ef10bacc9e67651a43d`) is not `>= 1.0.0` and Octane outputs this warning:

> WARN  Your FrankenPHP binary version (b2e92a673a8a5e1c285b6ef10bacc9e67651a43d) may be incompatible with Octane.
> 
>  Should Octane download the latest FrankenPHP binary version for your operating system? (yes/no) [yes]:

This makes it so Octane cannot be started by a supervisor process, even when using non-interactive or quiet flags (the default is `yes`, which keeps re-downloading the binary)

### Steps To Reproduce

1. Install a Laravel project with Octane on an Ubuntu server
2. Run `octane:install` and choose `frankenphp`
3. Let Octane download the binary
4. Try to run `octane:start`

nunomaduro:
Waiting for @dunglas to tag an Linux Build that displays the version like so:

```
FrankenPHP 1.0.1 PHP 8.3.0 Caddy v2.7.6 h1:w0NymbG2m9PcvKWsrXO6EEkY9Ru4FJK8uQbYcev1p3A=
```

nunomaduro:
Fixed.

tebbbb:
Using the latest laravel version as of today.

WARN  Your FrankenPHP binary version (1.0.3) may be incompatible with Octane.

PR #779, opened 2023-12-20
[2.x] Improves version fetching

nunomaduro:
This pull request improves the version fetching, as apparently the builds in different OSs display the version differently. So, using a `build-info` command is a better, more reliable way.

Fixes https://github.com/laravel/octane/issues/774, https://github.com/laravel/octane/issues/771.


cwhite92:
Nice!

PR #817, opened 2024-01-18
Listen Caddy admin port on all interfaces

smortexa:
Currently, Caddy admin APIs are not accessible from outside of container. This PR allows accessing Caddy admin APIs inside a container by listening on all interfaces.

### DIFF
--- a/src/Commands/Concerns/InteractsWithIO.php
+++ b/src/Commands/Concerns/InteractsWithIO.php
@@ -40,6 +40,7 @@ trait InteractsWithIO
         'write error',
         'unable to determine directory for user configuration; falling back to current directory',
         '$HOME environment variable is empty',
+        'admin endpoint on open interface; host checking disabled',
     ];
 
     /**
--- a/src/Commands/stubs/Caddyfile
+++ b/src/Commands/stubs/Caddyfile
@@ -1,7 +1,7 @@
 {
 	{$CADDY_GLOBAL_OPTIONS}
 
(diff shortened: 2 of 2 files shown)


francislavoie:
No, this is a very bad idea.

The admin endpoint should never be publicly accessible. Otherwise, an attacker could change your server's config and make it do whatever they want.

Many servers don't have firewalls blocking all ports by default, so listening to all interfaces here would be a huge security vulnerability (essentially a guarantee that your server will be compromised).

What you should do if you need to access the admin API is to change your Caddyfile yourself (copy it and modify it). Or alternatively you can use an SSH tunnel to access it (which can mimic the client being on the same machine).

mholt:
If the admin endpoint is opened up publicly, you should at least configure mTLS authentication to protect it.

Also, and this is way out in left field, but I believe using `localhost` as the interface implies a DNS lookup -- if the DNS resolver is out of your control then an attacker could potentially set it to a public interface. I usually prefer to hard-code the IP address here instead.

mholt:
If the admin endpoint is opened up publicly, you should at least configure mTLS authentication to protect it.

smortexa:
So, I think I probably should communicate with Caddy admin APIs (such as `/metrics`) through Laravel.

francislavoie:
Or use the `metrics` directive to expose it on any site in Caddy: https://caddyserver.com/docs/caddyfile/directives/metrics

PR #819, opened 2024-01-19
Fix Swoole in-memory timer table size

miguilimzero:
The current implementation of the timer table starts to throw some insufficient memory errors when there are a lot of workers ~200 and when all those workers are processing requests at the same time (as reported at https://github.com/laravel/octane/issues/648).

The https://github.com/laravel/octane/pull/650 PR introduced the `max_timer_table_size` undocumented config value so we can control the timer table size manually to avoid this problem.

However, ensuring the $size passed to the table creation will guarantee the desired number of rows (with  https://github.com/laravel/octane/pull/818). We can now use the number of workers to define the timer table size and get rid of the undocumented manual parameter.

### DIFF
--- a/bin/createSwooleTimerTable.php
+++ b/bin/createSwooleTimerTable.php
@@ -6,7 +6,7 @@ use Swoole\Table;
 require_once __DIR__.'/../src/Tables/TableFactory.php';
 
 if (($serverState['octaneConfig']['max_execution_time'] ?? 0) > 0) {
-    $timerTable = TableFactory::make($serverState['octaneConfig']['max_timer_table_size'] ?? 250);
+    $timerTable = TableFactory::make($serverState['workers']);
 
     $timerTable->column('worker_pid', Table::TYPE_INT);
     $timerTable->column('time', Table::TYPE_INT);


taylorotwell:
I don't understand why we wouldn't just document the configuration option instead?

miguilimzero:
@taylorotwell The timer table is internally used to check how much time a worker is processing a request. The timer table has a hard-coded size of 250, which starts to cause some trouble with a lot of workers processing many requests simultaneously, so the undocumented param was added to control the size manually.

There is no need to control the timer table size manually, as it will always need to be the size of the amount of available workers. With this PR, everything will just work out of the box.

This PR will not affect who is using this `max_timer_table_size` since everything now will just work. There is no custom need to set this table size manually.

miguilimzero:
The parameter would be on the config file, but it was removed before merging the original PR https://github.com/laravel/octane/pull/650/commits/dd3263b22e9f03d05bf51703d76ebfbc9e46c839. 

 I noted that the config param comment was wrong; it is not about max_execution_time but the number of workers processing requests simultaneously. The person who submitted the PR tried to solve the problem by adding the parameter and increasing it with a higher number when something goes wrong.
 
 **By setting the size equal to the number of workers, we are solving this problem in the correct way.**

PR #821, opened 2024-01-19
[2.x] Flushes `Once::class` on Operation Terminated

nunomaduro:
This pull request flushes the `Once::class` on the `OperationTerminated` event. This event is fired when:

- Request is handled.
- Tick is handled.
- Task is handled.

This pull request can be merged before or after : https://github.com/laravel/framework/pull/49744.


taylorotwell:
@nunomaduro I'm going to leave this on OperationTerminated instead of RequestTerminated.

PR #822, opened 2024-01-22
Add octane.roadrunner.config_version settings to roadrunner start binary

ValikoDorodnov:
For more comfortable developing i add an opportunity to set RoadRunner version via config.
 
 UPD
 
Also add:
- logs.mode
- logs.output
- logs.encoding


taylorotwell:

ValikoDorodnov:
@taylorotwell 
Thank you for your feedback, Taylor!

For this implementation, I used the same approach that is used on line 
```
src/Commands/StartRoadRunnerCommand.php:88
```

The function config() is standard for the Laravel library.

I think developers need to give the ability to manage parameters:
- version 
- logs.mode 
- logs.output 
- logs.encoding

I checked the build functionality, and it seems to be in order.

PR #853, opened 2024-03-15
Fix #851: When docker stop container, db connections still persists

### LINKED ISSUE #851: Octane doesn't drop connections after docker container stopped
### Octane Version

latest

### Laravel Version

11.0.5

### PHP Version

8.3

Open Swoole

### Server Version

docker latest

### Database Driver & Version

Mysql 8.3.0

### Description

We have a setup with docker + jenkins. When docker stops and starts a new container, the MySQL connection from the old container is still working.

After some time and several reboots, we use up all max_connections in the mysql container.

How i can handle this out?

By the way, `php artisan octane:start --workers=2 --task-workers=2 --host=0.0.0.0` should spawn 4 connections, i believe, but in my test, it spawn ~10

We can play with global variable in MySql, such as wait_timeout to force it close connection, but i want to discuss some other solutions, if such exist

### Steps To Reproduce

1. Setup mysql container via docker
2. Setup php container with `openswoole` and run it through `php artisan octane:start --workers=2 --task-workers=2 --host=0.0.0.0`
3. Use docker stop, rm and then up a new container.
4. Check PROCESSLIST in mysql, there will be old and new one connections.

amondar-libs:
fix for https://github.com/laravel/octane/issues/851

Problem:

When we run octane+openswoole in docker container we can't use CTRL+C combination to stop the server. 
In docker case, container operation system will be ripped during stop phase. In common cases, docker send SIGTERM signal into bash/sh terminal, that is an owner of the container. But, terminal SIGTERM not propagated into "octane" commands. 

Solution:

We should trap SIGHUP (terminal connection disrupted) signal, because only this signal will be sent into "octane" command, when terminal window closed. That can be easily tested on local machine: open OS task manager, find php stack and run "octane:start --server=swoole", and  close the terminal process. You will see, that php processes still there. But if we trap SIGHUP signal, octane command will receive the signal to stop the server.

That's it

### DIFF
--- a/src/Commands/Concerns/InteractsWithServers.php
+++ b/src/Commands/Concerns/InteractsWithServers.php
@@ -149,7 +149,7 @@ trait InteractsWithServers
      */
     public function getSubscribedSignals(): array
     {
-        return [SIGINT, SIGTERM];
+        return [SIGINT, SIGTERM, SIGHUP];
     }
 
     /**


driesvints:
@Focusnik please add a thorough description here and not just link to an issue.

amondar-libs:
@driesvints done

amondar-libs:
Also, i figured out some notice for docs mby. When we starting container in docker, we also make a volume for the logs folder. Octane creates server configuration file in the logs folder. So, when we scale docker container into >1 replicas with mounted logs directory we will have json file with pids for latest replica, that wrote into it. Developers must be careful with that and change state_file config value to some none mounted folder, so file will be unique per replica

nunomaduro (approved):
I've literally no idea if this will be a regression in the future. I think no; so will propose we merge this @taylorotwell and stay vigilant if anyone complains.

PR #854, opened 2024-03-15
[2.x] Closes monolog handlers by event listener if the worker stops

### LINKED ISSUE #850: Octane does not call close() on Monolog Handler
### Octane Version

2.3.4

### Laravel Version

10.46.0

### PHP Version

8.1.27

Swoole

### Server Version

5.1.2

### Database Driver & Version

_No response_

### Description

Laravel uses Monolog for logging, some monolog handler have a important `close()` function.
Octane does never call this function! In normal laravel applications, monolog calls close by itself via `__destruct()`.
It seams like `__destruct()` is never called in Swoole and it could be sensible to call `close()` after every request.

Taking into account that there is currently only one workaround to write performant logs: https://github.com/laravel/octane/issues/724#issuecomment-1649617729
In this case the `ProcessHandler::close()` is never called and laravel will spawn endless `cat` processes.

### Steps To Reproduce

1. Create an endpoint that writes information to the Log facade
2. Create a log driver that requires an `close()` call, like in https://github.com/laravel/octane/issues/724#issuecomment-1649617729
3. Run linux `ps` and see that for every request a `cat` process was created but not closed.

NiroDeveloper:
Fix for #850 
Laravel uses Monolog for logging, some monolog handler have a important close() function.
Octane does never call this function! In normal laravel applications, monolog calls close by itself via __destruct().
It seams like __destruct() is never called in Swoole, so close the handler manually per event.

### DIFF
--- a/config/octane.php
+++ b/config/octane.php
@@ -11,6 +11,7 @@ use Laravel\Octane\Events\TickTerminated;
 use Laravel\Octane\Events\WorkerErrorOccurred;
 use Laravel\Octane\Events\WorkerStarting;
 use Laravel\Octane\Events\WorkerStopping;
+use Laravel\Octane\Listeners\CloseMonologHandlers;
 use Laravel\Octane\Listeners\CollectGarbage;
 use Laravel\Octane\Listeners\DisconnectFromDatabases;
 use Laravel\Octane\Listeners\EnsureUploadedFilesAreValid;
@@ -114,6 +115,7 @@ return [
         ],
 
         WorkerStopping::class => [
+            CloseMonologHandlers::class,
             //
         ],
     ],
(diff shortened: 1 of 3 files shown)


taylorotwell:
Drafting until @nunomaduro can review.

nunomaduro (approved):

taylorotwell:
@NiroDeveloper how did you decide it should be placed on WorkerStopping? Why not OperationTerminated?

NiroDeveloper:
The handler should be reused for every request / operation, so it should just call `close()` at the end of the worker process.
There is already a `FlushMonologState` listener that calls `reset()` on the handler between every request, this is the correct behavior as descripted by monolog.

taylorotwell:
Thanks @NiroDeveloper

PR #855, opened 2024-03-15
Fix #838: The Symfony VarDumper dumps correctly when in HTML context

### LINKED ISSUE #838: dump() only works in cli
### Octane Version

2.23

### Laravel Version

10.43

### PHP Version

8.3.2

Swoole

### Server Version

5.1

### Database Driver & Version

_No response_

### Description

I've noticed that using `dump() `or `@dump` does not work with Laravel Octane when we are dumping a response for a browser. The `Symfony\Component\VarDumper\VarDumper` seems to explicitly check if `PHP_SAPI ` is 'cli' when dumping a variable. Since with Octane we're always in cli context, we're never using the `HtmlDumper`.
I found out that this issue can be circumvented by explicitly setting the handler for the `VarDumper ` like this when booting the app:

```
VarDumper::setHandler(function ($var, ?string $label = null) {
            $cloner = new VarCloner();
            $cloner->addCasters(ReflectionCaster::UNSET_CLOSURE_FILE_INFO);
            $var = $cloner->cloneVar($var);
            if (null !== $label) {
                $var = $var->withContext(['label' => $label]);
            }
            $dumper = app()->runningInConsole() ? new CliDumper() : new HtmlDumper();
            $dumper->dump($var);
        });

```

Ideally Octane would already handle this itself, unless I'm missing something.

### Steps To Reproduce

Start an octane server. Create a route that renders a blade view. Use `@dump($someVariable)` in the blade view. dump() will display nothing.

Add the VarDumper::setHandler command from above into the boot() method of your AppServiceProvider. dump() will now properly display $someVariable in the browser

AlliBalliBaba:
Fixes [#838](https://github.com/laravel/octane/issues/838)

When using `dump()` or `@dump` in Octane nothing is displayed when we're in the context of a request.
One would expect a dump in form of HTML like it's the case in regular Laravel

This commit fixes the Symfony `VarDumper` so it detects whether we're in console or not by using `app()->runningInConsole()` instead of looking at the value of `PHP_SAPI`.

The test I wrote checks that Blade explicity renders a dumped variable. It is unfortunate that I have to explicitly set `$_ENV['APP_RUNNING_IN_CONSOLE']` in the test and unset it afterwards. It probably would be better to generally mock `app()->runningInConsole()` so it returns false for all tests. In the scope of this fix I did not want to touch other tests though.

Cheers

### DIFF
--- a/config/octane.php
+++ b/config/octane.php
@@ -68,7 +68,7 @@ return [
         WorkerStarting::class => [
             EnsureUploadedFilesAreValid::class,
             EnsureUploadedFilesCanBeMoved::class,
-            EnsureTheCorrectVarDumperIsUsed::class
+            EnsureTheCorrectVarDumperIsUsed::class,
         ],
 
         RequestReceived::class => [
--- a/config/octane.php
+++ b/config/octane.php
@@ -13,6 +13,7 @@ use Laravel\Octane\Events\WorkerStarting;
 use Laravel\Octane\Events\WorkerStopping;
 use Laravel\Octane\Listeners\CollectGarbage;
 use Laravel\Octane\Listeners\DisconnectFromDatabases;
+use Laravel\Octane\Listeners\EnsureTheCorrectVarDumperIsUsed;
 use Laravel\Octane\Listeners\EnsureUploadedFilesAreValid;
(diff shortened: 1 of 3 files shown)


taylorotwell:

Issue #809, opened 2024-01-13
UrlGenerator warming up issue


gazzoy:
### Octane Version

2.2.7

### Laravel Version

10.40.0

### PHP Version

8.2.14

### What server type are you using?

Swoole

### Server Version

5.1.1

### Description

Same as #784, #748.

When using `Route::bind()` and `URL::defaults()` with Octane, it seems that `UrlGenerator` keeps the same instance of `RouteUrlGenerator`, and so the defaults are cached between requests.

Especially, I believe the issue happens when we use `App\Http\Middleware\Authenticate` to handle login sessions.

### Steps To Reproduce

I created a public repo [gazzoy/laravel-octane-urlgenerator-issue](https://github.com/gazzoy/laravel-octane-urlgenerator-issue) with a fresh Laravel v10.40.0, Octane v2.2.7(swoole), Sail v1.27.0(php8.2), and the smallest code to reproduce the issue.

1. clone [gazzoy/laravel-octane-urlgenerator-issue](https://github.com/gazzoy/laravel-octane-urlgenerator-issue)
2. put an attached [.env.txt](https://github.com/laravel/octane/files/13927525/default.env.txt) on your local path as `gazzoy/laravel-octane-urlgenerator-issue/.env`.
3. run below commands.
```shell
$ cd gazzoy/laravel-octane-urlgenerator-issue
$ docker run --rm \
    -u "$(id -u):$(id -g)" \
    -v "$(pwd):/var/www/html" \
    -w /var/www/html \
    laravelsail/php82-composer:latest \
    composer install --ignore-platform-reqs
$ ./vendor/bin/sail build --no-cache
$ ./vendor/bin/sail up -d
$ ./vendor/bin/sail artisan migrate:refresh --seed
```
4. open your browser with:
`http://localhost/sample/paul/login`
It should show a page with a text "login page for paul" which is expected. 

5. open below url on the same browser:
`http://localhost/sample/john/home`
It redirects to `http://localhost/sample/paul/login` which doesn't make sense to me. Because, as of I understood, if `Route::bind()` and `URL::defaults()` work fine, we should be redirected to `http://localhost/sample/john/login`. Thefore, I believe it seems URL::defaults() keeps it states?

Having said that, I'm not too sure its actually Octane issue, so please let me know if this is not the case.

[...]

PR #869, opened 2024-04-18
Add failing test for UrlGenerator warming up issue

gazzoy:
This aims to add failing test for https://github.com/laravel/octane/issues/809. To recap, the issue is when using Route::bind() and URL::defaults() with Octane, it seems that UrlGenerator keeps the same instance of RouteUrlGenerator, and so the defaults are cached between requests.

There is another problem in this test which I got `Route [login] not defined`.  It doesn't really make sense to me since I put `->name('login')` in route definition, but I might be wrong.

Please let me know if you have any questions.


taylorotwell:
We don't merge failing tests. There is no action to take here. We accept fixes via PRs.

PR #870, opened 2024-04-18
fix: windows install error, don`t need to chmod (rr.exe)

seth-shi:
fix: windows install error, don`t need to chmod (rr.exe)
```shell
ErrorException 

 chmod(): No such file or directory
```


taylorotwell:
Octane doesn't support Windows.

PR #874, opened 2024-04-18
fix: use capture to populate global request with json array

krowinski:
Request::capture is using createFromBase that checks is request is json and replace request with json array

``` 
       if ($newRequest->isJson()) {
            $newRequest->request = $newRequest->json();
        } 
```

and later when in controller you can use request->get('some-json-key') to get value as its a default laravel behavior [check index.php:53]

### DIFF
--- a/src/FrankenPhp/FrankenPhpClient.php
+++ b/src/FrankenPhp/FrankenPhpClient.php
@@ -19,7 +19,7 @@ class FrankenPhpClient implements Client
     public function marshalRequest(RequestContext $context): array
     {
         return [
-            Request::createFromGlobals(),
+            Request::capture(),
             $context,
         ];
     }


taylorotwell:
No tests and no real explanation of how to recreate the issue this is supposed to fix.

krowinski:
@taylorotwell 

Create route

```
Route::post('/test', static function (Request $request) {
    return new JsonResponse($request->get('key'));
});
```

Post json 

```
curl --location 'localhost/test/' \                                                                                                                               
--header 'Content-Type: application/json' \
--data '{"key":"value"}'
```

Expected `"value"`
Current  "{}"


Laravel is using ```$response = $kernel->handle($request = Request::capture())->send();``` in index.php but FrankenPhpClient is just doing Request::createFromGlobals() and its not passing by \Illuminate\Http\Request::createFromBase that populate request using incoming json values using this code fragment.

```     
       if ($newRequest->isJson()) {
            $newRequest->request = $newRequest->json();
        }
```

PR #878, opened 2024-04-26
Boot RoadRunner worker before first request

daannet:
The RoadRunner worker is currently only initialized and booted once a request is made. 

With this change, the worker is already booted when the first request is received. As a result, the first request will be executed much faster because the configured 'warm' services are already booted. 

Before commit d9b2a33 this was already the case. In 6b4db3a the while & try is flipped.


taylorotwell:
With this change if there is an error booting the framework it can be hidden from the user.

daannet:
@taylorotwell Where else should any boot errors be visible than the Roadrunner logs?

Issue #887, opened 2024-05-12
Memory Leak in ViewServiceProvider


AlliBalliBaba:
### Octane Version

2.3.10

### Laravel Version

11.4.0

### PHP Version

8.3.7

### What server type are you using?

FrankenPHP

### Server Version

1.1.4

### Description

While doing some tests with Laravel Octane and the FrankenPHP/Swoole runtimes I noticed that the application was getting slower as time went on, especially when setting a big number for `max-requests`.
After digging around I found that it only happens when a blade file is rendered. I found out the `ViewServiceProvider` is the issue since it's holding a stale version of the `$app` container.

More specifically it's line 166-168 where a terminating callback gets added to the app every time a new Blade compiler is resolved (see image).

<img width="555" alt="bug" src="https://github.com/laravel/octane/assets/45872305/e2581eae-99ac-447d-978f-e0bd57915e79">

For every request a new callback gets added to the original app instance. At some point there are thousands of callbacks in the `terminatingCallbacks[]` array of the original `$app` , which slows down the application and is a **memory leak**.

It suspect Octane is only giving the new `$app `instance to the `ViewFactory ` but not to the `ViewServiceProvider`

### Steps To Reproduce

Set up a clean Laravel installation with Laravel Octane. 
install dependencies with composer install
install any runtime with `php artisan octane:install`, I'll choose FrankenPHP

Go into the vendor folder of your application and modify the `Illuminate\Foundation\Application` like this to dump out the amount of registered callbacks:

```
public function terminating($callback)
    {
         $this->terminatingCallbacks[] = $callback;
+       Log::info(count($this->terminatingCallbacks) . " callbacks registered");

        return $this;
    }
```

start octane with 1 worker for simplicity:
`php artisan octane:start --server=frankenphp --host=yourhost --port=80 --max-requests=1000' --workers=1 --admin-port=2019
`

visit the default home page `/home` in the browser multiple times. In your `laravel.log` you can now see that the number of callbacks increases on each request.

driesvints:
Thanks for reporting this @AlliBalliBaba. Would appreciate a PR if anyone can figure one out.

AlliBalliBaba:
I think the main issue here and with possible other memory leaks is that Octane changes the reference to the application container on each request and then attempts to update all references to the old instance. A safer approach would be to not clone the container,  but reset it to an earlier state instead. 

Here is what I mean:

[...]

PR #888, opened 2024-05-13
Fixing memory leaks that stem from the cloned application instance

AlliBalliBaba:
This pull request seeks to fix memory leaks like the one in [#887](https://github.com/laravel/octane/issues/887).
One big downside about Octane's approach of cloning a new application instance on each request is that it's not really possible to track down all references to the original application instance. This can lead to memory leaks that are hard to detect and incompatibilities with other Laravel libraries.

In this pull request I put forward a new approach that seeks to instead keep the same application instance but reset it to a previous state:

octane drawio

This can be achieved pretty elegantly by creating an `ApplicationSnapshot` that extends the base Application and stores its initial state. Since PHP objects have access to the protected scope of other objects of the same class, the Snapshot is able to reset all Application properties via `get_object_vars()`. This ends up doing the same thing as `clone`, but without creating a new instance.

I tested this approach with the Swoole and FrankenPHP runtimes and it indeed fixes the memory leak in [#887](https://github.com/laravel/octane/issues/887) and should generally make integrating Octane with other `ServiceProviders `easier.

The main changes are in the `Worker.php` class and the newly added `ApplicationSnapshot.php` class. All other other changes mainly remove code that becomes redundant through this PR and adjusts tests that expect the application instance to change.

This is just an initial draft. There is probably potential for more code removal and the approach should be thoroughly tested with other first party packages.

### DIFF
--- a/src/Listeners/GiveNewApplicationInstanceToViewFactory.php
+++ b/src/Listeners/GiveNewApplicationInstanceToViewFactory.php
@@ -24,7 +24,7 @@ class GiveNewApplicationInstanceToViewFactory
         });
 
         $provider = $event->sandbox->getProvider(ViewServiceProvider::class);
-        if($provider && method_exists($provider, 'setApplication')){
+        if ($provider && method_exists($provider, 'setApplication')){
             $provider->setApplication($event->sandbox);
         }
     }
--- a/src/Listeners/GiveNewApplicationInstanceToViewFactory.php
+++ b/src/Listeners/GiveNewApplicationInstanceToViewFactory.php
@@ -24,7 +24,7 @@ class GiveNewApplicationInstanceToViewFactory
         });
 
         $provider = $event->sandbox->getProvider(ViewServiceProvider::class);
-        if ($provider && method_exists($provider, 'setApplication')){
+        if ($provider && method_exists($provider, 'setApplication')) {
             $provider->setApplication($event->sandbox);
         }
     }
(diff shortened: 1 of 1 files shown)


driesvints:
Marking this as ready for review so it can be reviewed. Thanks!

AlliBalliBaba:
Something else that would make this even more safe is removing all access to the initial app state in Octane's event listeners. The flushing of scoped instances and so on would then happen right before making the snapshot instead of after each request. That would 100% ensure that only one application instance ever exists and that its state is exactly the same before each request.

Doing so might break custom event listeners people have added to the `octane php` config though.

taylorotwell:
This is just an extremely risky PR for us to merge on a patch release. Can we instead focus on **just** fixing the ViewServiceProvider problem first?

AlliBalliBaba:
I agree, this should probably be refined more for a minor or major.
Fixing the leak might not be possible though without adding an `$app` setter in the `ViewServiceProvider` class in `laravel/framework`.
Would the strategy there be to first make a pull request in `laravel/framework` and then make the pull request here?

driesvints:
@AlliBalliBaba yes please

AlliBalliBaba:
I've reset this PR to only specifically fix the #887 memory leak. The PR is waiting for [another PR in laravel/framework](https://github.com/laravel/framework/pull/51443). It won't break older versions since it checks if the `method_exists`.

I will revisit the Snapshot approach in a future PR since I think it will make Octane more maintainable in the long run.

nunomaduro:
Fixed here: https://github.com/laravel/framework/pull/51450.

PR #891, opened 2024-05-14
Hide errors on production

chu121su12:
Let errors be hidden on APP_ENV=production


taylorotwell:
Needs more explanation.

PR #892, opened 2024-05-14
Allow debug in CADDY_GLOBAL_OPTIONS

chu121su12:
This PR let `debug` option to be appended in CADDY_GLOBAL_OPTIONS. See https://caddyserver.com/docs/caddyfile/options#debug

### DIFF
--- a/src/Commands/StartFrankenPhpCommand.php
+++ b/src/Commands/StartFrankenPhpCommand.php
@@ -95,7 +95,11 @@ class StartFrankenPhpCommand extends Command implements SignalableCommandInterfa
             'LARAVEL_OCTANE' => 1,
             'MAX_REQUESTS' => $this->option('max-requests'),
             'REQUEST_MAX_EXECUTION_TIME' => $this->maxExecutionTime(),
-            'CADDY_GLOBAL_OPTIONS' => ($https && $this->option('http-redirect')) ? '' : 'auto_https disable_redirects',
+            'CADDY_GLOBAL_OPTIONS' => trim(sprintf(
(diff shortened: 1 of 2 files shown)


src/Commands/StartFrankenPhpCommand.php
```diff
@@ -95,7 +95,11 @@ public function handle(ServerProcessInspector $inspector, ServerStateFile $serve
             'LARAVEL_OCTANE' => 1,
             'MAX_REQUESTS' => $this->option('max-requests'),
             'REQUEST_MAX_EXECUTION_TIME' => $this->maxExecutionTime(),
-            'CADDY_GLOBAL_OPTIONS' => ($https && $this->option('http-redirect')) ? '' : 'auto_https disable_redirects',
+            'CADDY_GLOBAL_OPTIONS' => trim(sprintf(
+                '%s %s',
```
dunglas:
```suggestion
                "%s\n%s",
```

dunglas (commented):
Alternatively, to simplify the code, we could introduce more environment variables for these options (like `DEBUG` and `AUTO_HTTPS`).

chu121su12:
AUTO_HTTPS is already part of `http-redirect`. Added `server-debug` option only for FrankenPHP for now

taylorotwell:
I would actually like to see these become env vars otherwise we may end up with a lot of FrankenPHP specific CLI options?

driesvints:
Going to close this one now as it seems stale. Feel free to resend this once you've made the necessary changes, thanks.

Issue #893, opened 2024-05-16
Swoole - Access logs for non local/testing environments


tania-pets:
When using `swoole` server, the access logs are logged for local/testing environments only, see [here](https://github.com/laravel/octane/blob/2.x/src/Swoole/Handlers/OnWorkerStart.php#L97), and I can't find a way to override that.

It's important to see those access logs for production environments and it would be great to have a config variable to toggle the access logs.

Thank you

driesvints:
Thanks. It seems you already sent in a PR so let's see how it goes.

tania-pets:
> Thanks. It seems you already sent in a PR so let's see how it goes.

@driesvints Can we reopen this one, since my pr didn't go through, but didn't get any feedback for future plans?

driesvints:
@tania-pets it seems Taylor doesn't wants to change the behaviour for this one right now sorry so we're going to leave this one be.

PR #894, opened 2024-05-16
Add enable_access_logs boolean config for swoole

tania-pets:
Adds `octane.swoole.enable_access_log` config variable to toggle access logs for swoole server.
See https://github.com/laravel/octane/issues/893

~~**It defaults to `true`, that means it will enable the logs for non-local environments if it's not set to false explicitly in the config file.~~

**It defaults to `false`, that means it won't enable the logs for non-local environments, but it will stop logging for local/testing

### DIFF
--- a/src/Swoole/Handlers/OnWorkerStart.php
+++ b/src/Swoole/Handlers/OnWorkerStart.php
@@ -97,7 +97,7 @@ class OnWorkerStart
             if (! config('octane.swoole.enable_access_logs', true)) {
                 return;
             }
-            
+
             Stream::request(
                 $request->getMethod(),
                 $request->fullUrl(),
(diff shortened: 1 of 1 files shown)


taylorotwell:
Since it defaults to true would this also enable it in production for existing apps?

tania-pets:
> Since it defaults to true would this also enable it in production for existing apps?

@taylorotwell Exactly, yes.
In order to maintain the current behaviour, we could do it like:

```
if (! $sandbox->environment('local', 'testing')) {
    if (! config('octane.swoole.enable_access_logs', false)) {
        return;
    }
}
```            

That way the logs will be always enabled for local/testing and the config will control non-local only. The config variable name will be misleading though, because it will have no effect on local/testing.
We could rename it to something like `enable_non_local_access_logs` or something, but I think it's much better to be able to control this per environment, regardless if it's local or not (There can be multiple non local environments, and each should be configurable)

But I do get the concern about changing production behaviour.

So maybe it's better to just default it to false:
```
if (! config('octane.swoole.enable_access_logs', false)) {
    return;
}
```
This ^ will switch off the logs for local/testing, but changing such a local behaviour is less concerning than changing production. PR updated.

driesvints:
@tania-pets just marking this as ready again so Taylor sees your response

taylorotwell:
Sorry, this also changes the behavior as now they won't be enabled on local dev.

PR #896, opened 2024-05-20
UrlGenerator Sandbox

### LINKED ISSUE #809: UrlGenerator warming up issue

dbpolito:
Currently we're sending the same instance of URL across all requests, when changing `defaults` or `forceRootUrl`, the change is happening across all requests...

This should also fix https://github.com/laravel/octane/issues/809

### DIFF
--- a/src/Concerns/ProvidesDefaultConfigurationOptions.php
+++ b/src/Concerns/ProvidesDefaultConfigurationOptions.php
@@ -28,6 +28,7 @@ trait ProvidesDefaultConfigurationOptions
     {
         return [
             \Laravel\Octane\Listeners\CreateConfigurationSandbox::class,
+            \Laravel\Octane\Listeners\CreateUrlGeneratorSandbox::class,
             \Laravel\Octane\Listeners\GiveNewApplicationInstanceToAuthorizationGate::class,
             \Laravel\Octane\Listeners\GiveNewApplicationInstanceToBroadcastManager::class,
             \Laravel\Octane\Listeners\GiveNewApplicationInstanceToDatabaseManager::class,
--- a/src/Listeners/CreateUrlGeneratorSandbox.php
+++ b/src/Listeners/CreateUrlGeneratorSandbox.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Laravel\Octane\Listeners;
+
+class CreateUrlGeneratorSandbox
+{
+    /**
+     * Handle the event.
+     *
+     * @param  mixed  $event
+     */
+    public function handle($event): void
+    {
+        $event->sandbox->instance('url', clone $event->sandbox['url']);
+    }
+}
--- a/src/OctaneServiceProvider.php
+++ b/src/OctaneServiceProvider.php
@@ -123,6 +123,7 @@ class OctaneServiceProvider extends ServiceProvider
     {
         $this->app->singleton(Listeners\CollectGarbage::class);
         $this->app->singleton(Listeners\CreateConfigurationSandbox::class);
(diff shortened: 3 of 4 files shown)


taylorotwell:
I wonder if this would be a breaking change, especially around the `defaults` stuff. If someone set a default for localization purposes for example and it was lost on subsequent requests it seems like that could be a problem?

dbpolito:
Well it may depend on where you call it... But if you call it on on middleware where it's recommended it shouldn't have any impact as in theory you will always be setting it...

But tbh, i think the potential problems we currently have is probably bigger than the people that may be impacted (which is very minimum i think)... 🤔

PR #898, opened 2024-05-24
Add `--admin-host` argument to FrankenPHP Command

alextricity25:
As a user of Octane using FrankenPHP, I would like to override the host binding of the Caddy API. This will allow users to do things like scraping metrics from the API server (/metrics) from a remote host.


src/Commands/StartFrankenPhpCommand.php
```diff
@@ -238,6 +241,20 @@ protected function adminPort()
         });
     }
 
+    /**
+     * Get the host the admin API should be available on.
+     *
+     * @return string
+     */
+    protected function adminHost()
```
taylorotwell: Can you just make a `localhost` the default value of the option so we don't need this method?

alextricity25:
Thanks @taylorotwell! Great observation! I've made your suggested edits in the latest commit.

PR #899, opened 2024-05-25
Give New Application to Routes

dbpolito:
Routes instances are still referencing the old container...

I was hitting this error in a filament / livewire app, more specifically here: https://github.com/livewire/livewire/blob/main/src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php#L149

Using cached routes does not hit this problem, that's why i'm only doing to uncached routes.

### DIFF
--- a/src/Listeners/GiveNewApplicationInstanceToRouter.php
+++ b/src/Listeners/GiveNewApplicationInstanceToRouter.php
@@ -2,6 +2,8 @@
 
 namespace Laravel\Octane\Listeners;
 
+use Illuminate\Routing\RouteCollection;
+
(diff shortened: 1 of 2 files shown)


taylorotwell:
What actually goes wrong?

dbpolito:
It's a very specific error, if a have a `Controller` with some injection on `__construct`, it throws this error:

```
Target [xxxContract] is not instantiable while building [xxxController, xxxRepository].
```

As the route is already loaded but referencing the old container, it's trying to inject using the wrong container, which probably got cleared.

The error happens on the livewire action, at the livewire.update ajax.

PR #910, opened 2024-06-18
Refreshes log manager container

### LINKED ISSUE #907: Context gets cleared before logging
### Octane Version

2.3.7

### Laravel Version

11.10

### PHP Version

8.3

Swoole

### Server Version

AWS

### Database Driver & Version

MySQL 8

### Description

While using Laravel Octane and Context variables, it gets cleared before the log entry is created. I can successfully access my data by `Context::get('name')` in the app, but it's not being included in logs.

I tried to run this application without Octane (`php artisan serve`), and the context is working as expected.

### Steps To Reproduce

Create a test route in `web.php`: (As shown [here](https://youtu.be/q8kyb6uArx0?t=23))

```php
use Illuminate\Support\Facades\Context;
use Illuminate\Support\Facades\Log;

Route::get('test', function () {
    Context::add('global_context', 'value');

    Log::info('test', ['local_data' => 'local']);

    return response('test');
});

```

In `laravel.log` file, you will get only `local_data`.

```
[2024-06-15 10:45:51] local.INFO: test {"local_data":true}
```

timacdonald:
This PR ensures that the log manager gets its container refreshed. Context is not being appended to log files as expected due to the stale container.

This is required because we do not eagerly resolve context from the container, so the parent container will never a context repository instance resolved (they belong to the sandbox application).

https://github.com/laravel/framework/blob/02a0b5f6f83b13ea750f959364aefa324441a354/src/Illuminate/Log/LogManager.php#L142-L149

fixes https://github.com/laravel/octane/issues/907

### DIFF
--- a/src/Concerns/ProvidesDefaultConfigurationOptions.php
+++ b/src/Concerns/ProvidesDefaultConfigurationOptions.php
@@ -35,6 +35,7 @@ trait ProvidesDefaultConfigurationOptions
             \Laravel\Octane\Listeners\GiveNewApplicationInstanceToDatabaseSessionHandler::class,
             \Laravel\Octane\Listeners\GiveNewApplicationInstanceToFilesystemManager::class,
             \Laravel\Octane\Listeners\GiveNewApplicationInstanceToHttpKernel::class,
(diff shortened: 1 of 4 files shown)


driesvints:
Drafting this until we can fix the tests.

PR #912, opened 2024-06-18
Rollback existing transactions after request ended

### LINKED ISSUE #909: Rollback any existing transactions after request ended
Consider the following code:

```
DB::beginTransaction();
// Lots of inserts
// Insert user
// Insert/add roles
if (someCondition()) {
   abort(403);
)
DB::commit();
```

If someCondition() is true the first time, then the transaction is not rolled back.
If the second time someCondition() is false then the previous transaction's queries are also committed.

I used to code with this in mind:

> When the script ends or when a connection is about to be closed, if you have an outstanding transaction, PDO will automatically roll it back. This is a safety measure to help avoid inconsistency in the cases where the script terminates unexpectedly--if you didn't explicitly commit the transaction, then it is assumed that something went awry, so the rollback is performed for the safety of your data.

[PHP PDO transactions](https://www.php.net/manual/en/pdo.transactions.php)
But since Octane never actually quits, the transaction is not rolled back.

So my suggestion would be to add an event to a listener for example `RequestTerminated` that rolls back any existing transaction.
So that unexpected rows aren't committed to the database.

vatsake:
> When the script ends or when a connection is about to be closed, if you have an outstanding transaction, PDO will automatically roll it back. This is a safety measure to help avoid inconsistency in the cases where the script terminates unexpectedly--if you didn't explicitly commit the transaction, then it is assumed that something went awry, so the rollback is performed for the safety of your data.

[PHP PDO transactions](https://www.php.net/manual/en/pdo.transactions.php)

Since Octane never actually quits, the transaction will still exist after the request has ended.

Consider the following code:

```
DB::beginTransaction();
User::create(['name' => $request->input('name')]);
if (someCondition()) {
   abort(403);
)
DB::commit();
```

If for some reason someCondition() is true, the request is aborted, but the transaction is not ended.
If the user tries again and someCondition() is false, then both the new and previous 'User' model is created.

Fixes
#909 

Unfortunately I don't know how to test this. Hopefully someone will add it.

### DIFF
--- a/config/octane.php
+++ b/config/octane.php
@@ -20,6 +20,7 @@ use Laravel\Octane\Listeners\FlushOnce;
 use Laravel\Octane\Listeners\FlushTemporaryContainerInstances;
 use Laravel\Octane\Listeners\FlushUploadedFiles;
 use Laravel\Octane\Listeners\ReportException;
+use Laravel\Octane\Listeners\RollbackActiveTransactions;
 use Laravel\Octane\Listeners\StopWorkerIfNecessary;
 use Laravel\Octane\Octane;
 
@@ -77,7 +78,7 @@ return [
         ],
 
         RequestHandled::class => [
-            //
+            RollbackActiveTransactions::class
         ],
 
         RequestTerminated::class => [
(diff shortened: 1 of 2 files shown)


taylorotwell:
I would add this to your own application if you need it.

driesvints:
Sorry @vatsake. Seems like we're not going to make this change in core atm.

Issue #928, opened 2024-07-17
Your FrankenPHP binary version (1.0.3) may be incompatible with Octane


tebbbb:
### Octane Version

1.0.3

### Laravel Version

11.8.0

### PHP Version

8.3

### What server type are you using?

FrankenPHP

### Server Version

1.0.3

### Description

Using the latest laravel version as of today.

WARN  Your FrankenPHP binary version (1.0.3) may be incompatible with Octane.

Using FROM dunglas/frankenphp:1.0-php8.3

```
FROM dunglas/frankenphp:1.0-php8.3

ARG DOCKER_WWWGROUP
ARG DOCKER_WWWUSER
ARG NODE_VERSION=20
ARG APP_ENV

WORKDIR /var/www/html

ENV DEBIAN_FRONTEND=noninteractive
ENV TZ=UTC

# Set timezone
RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone

# Update and install essential packages
RUN apt-get -y update --fix-missing && \
    apt-get upgrade -y && \
    apt-get install -y --no-install-recommends \
        apt-utils \
        apt-transport-https \
        curl \
        wget \
        nano \
        dialog \
        build-essential \
        git \
        gnupg2 \
        iputils-ping \
        libonig-dev \
        libsqlite3-dev \
        libsqlite3-0 \
        libzip-dev \
        libicu-dev \
        libxml2-dev \
        libxslt-dev \
        libssh2-1-dev \
        libssh2-1 \
        libcurl4-openssl-dev \
        pkg-config \
        openssl \
        zip \
        unzip \
        supervisor && \
    rm -rf /var/lib/apt/lists/*

# Install Node.js and Yarn
RUN curl -sL https://deb.nodesource.com/setup_$NODE_VERSION.x | bash - && \
    apt-get install -y --no-install-recommends nodejs && \
    npm install --global yarn && \
    rm -rf /var/lib/apt/lists/*

RUN pecl install xdebug

# Install PHP extensions
RUN install-php-extensions \
    pdo_sqlite \
    curl \
    zip \
    intl \
    mbstring \
    gettext \
    calendar \
    exif \
    pcntl \
    dom \
    xsl \
    simplexml \
    bcmath \
    sockets

# Clean up the image
RUN apt-get clean && apt-get autoremove -y

# Get latest Composer
COPY --from=composer:latest /usr/bin/composer /usr/bin/composer

# Create user and group
[...]

PR #931, opened 2024-07-24
Automatically disable FrankenPHP worker mode locally

Radiergummi:
This PR updates the FrankenPHP configuration to disable worker mode if the application is running in the `local` environment and the worker count is not specified explicitly (or set to `auto`).

### The problem
FrankenPHP's [Worker mode](https://frankenphp.dev/docs/worker/) keeps the application in memory, which does wonders for performance when running in production. Locally, however, it [causes problems](https://github.com/dunglas/frankenphp/issues/931) with XDebug, and does not reflect file changes immediately. The solution to the latter is configuring Octane to restart the worker after every request, thereby defeating most of the advantages of an application server in the first place.

### The solution
By only enabling worker mode if the application is running in a non-local environment, it will be served in a CGI-like manner, i.e. what the built-in PHP server (used in `artisan serve`) does—and exactly what you'd expect to happen: XDebug *just works*, and changes to files are reflected immediately.  
In production, worker mode will be enabled and all the benefits that come with it apply. We use the setup as provided in this PR in our production application and don't face any issues with it.

This solves https://github.com/dunglas/frankenphp/issues/931, and probably solves #928 too.


miguno:
@Radiergummi : I am running into the same issue as https://github.com/dunglas/frankenphp/issues/931.  What's the status of this PR?  Was the issue fixed elsewhere in the meantime?

Edit: It seems there is an updated PR at https://github.com/laravel/octane/pull/932.

miguno:
Ah, maybe there's an updated PR at https://github.com/laravel/octane/pull/932.

Radiergummi:
@miguno that one's on me—it took me some time to implement some required changes—sorry. As soon as the updates have been reviewed, this should be working!

miguno:
No worries, thanks a lot for your work on fixing this problem!

PR #939, opened 2024-08-02
fix: RoadRunner streaming through generators

### LINKED ISSUE #903: Streamed response gets buffered and returned in one chunk
### Octane Version

2.3.12

### Laravel Version

11.9.2

### PHP Version

8.3.0

Roadrunner

### Server Version

2024.1.2

### Database Driver & Version

_No response_

### Description

When streaming a response from a route behind Octane, the response gets buffered and arrives all at once instead of as it is generated.

Currently, `wire:stream` in Livewire is [completely broken](https://github.com/livewire/livewire/discussions/7798) when using Laravel Octane, but the issue is not specific to Livewire and my instructions below use native Laravel and vanilla JavaScript only.

### Steps To Reproduce

I have created a [minimal reproduction repository](https://github.com/danharrin/octane-streaming). Follow the instructions in the README to observe the behaviour of `artisan serve` as opposed to Octane. There is a small amount of JavaScript in `welcome.blade.php` that sends the request and outputs the chunks in real time.

Here is the response, note the `sleep(1)` calls to simulate an artificial delay in receiving the stream chunks. When using `artisan serve`, there is a second delay between each number appearing. When using Octane, there is a 5 second delay before all numbers appear at once.

```php
return response()->stream(function () {
    echo 1;

    ob_flush();
    flush();
    sleep(1);

    echo 2;

    ob_flush();
    flush();
    sleep(1);

    echo 3;

    ob_flush();
    flush();
    sleep(1);

    echo 4;

    ob_flush();
    flush();
    sleep(1);

    echo 5;
}, 200, [
    'Content-Type' => 'text/html; charset=utf-8;',
    'Cache-Control' => 'no-cache',
    'X-Accel-Buffering' => 'no',
]);
```

danharrin:
Fixes #903.

RoadRunner supports streamed responses, but not in the same way that other servers do. Instead out using the output buffer with `echo`, it supports using a `Generator` function as the response  callback to "yield" each chunk of the response.

Previous fixes have involved collecting the output buffer and returning it at once from the server. This is not streaming as all chunks are collected into memory and sent at once.

This implementation allows you to use a `Generator` as the callable in the streamed response. Each `yield` sends a chunk in the response. It is quite simple and just interacts with the underlying `HttpWorker` that is used by the Roadrunner client. The new code is only invoked if the streamed response function is a generator, else the old behaviour still persists.

```php
use Generator;

return response()->stream(function (): Generator {
    yield 1;

    sleep(1);

    yield 2;

    sleep(1);

    yield 3;

    sleep(1);

    yield 4;

    sleep(1);

    yield 5;
}, 200, [
    'Content-Type' => 'text/html; charset=utf-8;',
    'Cache-Control' => 'no-cache',
    'X-Accel-Buffering' => 'no',
]);
```

Note: we do not need to evaluate the function to know if we need to stream it in this way, we use reflection to do this ahead of time, ensuring that the function returns a `Generator` object.

I would like to thank @donnysim for his investigative work and initial solution which was adapted into this PR.

### DIFF
--- a/src/RoadRunner/RoadRunnerClient.php
+++ b/src/RoadRunner/RoadRunnerClient.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Octane\RoadRunner;
 
+use Generator;
 use Illuminate\Foundation\Application;
 use Illuminate\Http\Request;
 use Laravel\Octane\Contracts\Client;
@@ -10,6 +11,7 @@ use Laravel\Octane\MarshalsPsr7RequestsAndResponses;
 use Laravel\Octane\Octane;
 use Laravel\Octane\OctaneResponse;
 use Laravel\Octane\RequestContext;
+use ReflectionFunction;
 use Spiral\RoadRunner\Http\PSR7Worker;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
 use Symfony\Component\HttpFoundation\StreamedResponse;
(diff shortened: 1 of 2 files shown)


src/RoadRunner/RoadRunnerClient.php
```diff
@@ -47,6 +49,20 @@ public function respond(RequestContext $context, OctaneResponse $octaneResponse)
             );
         }
 
+        if (
+            ($octaneResponse->response instanceof StreamedResponse) &&
+            ($responseCallback = $octaneResponse->response->getCallback()) &&
+            ((new ReflectionFunction($responseCallback))->getReturnType()?->getName() === Generator::class)
```
donnysim, line 55: This could be:
```php
((new ReflectionFunction($responseCallback))->isGenerator()
```
a typehint would not be needed to know if it's a generator or not

donnysim:
I'm so sorry, I completely forgot about the whole thing and could've saved you some trouble if you were too busy 🙏. This looks good though 👍.

PR #946, opened 2024-09-10
Add keep-config option to install command

MatusBoa:
In our company, we are building docker image containing octane. One of the build step is installing roadrunner (`php artisan octane:install --server=roadrunner`), which is overwriting our existing configuration for octane.


taylorotwell:
Can you just commit Roadrunner into your repo?

MatusBoa:
@taylorotwell i would like to keep multi-arch support, meaning i need to download roadrunner binary within build for specific platform

PR #956, opened 2024-09-25
[2.x] Supports Laravel Prompts 0.2 and 0.3

webard:
`laravel/framework` introduces `laravel/prompts:^0.2`, but `laravel/octane` still requires `^0.1`.

This is just a simple change in composer.json. I’m not sure if Octane should support both versions or only `^0.2`, so I’ve left both versions for now.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "php": "^8.1.0",
         "laminas/laminas-diactoros": "^3.0",
         "laravel/framework": "^10.10.1|^11.0",
-        "laravel/prompts": "^0.1.24",
+        "laravel/prompts": "^0.1.24|^0.2",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -17,7 +17,7 @@
         "php": "^8.1.0",
         "laminas/laminas-diactoros": "^3.0",
         "laravel/framework": "^10.10.1|^11.0",
-        "laravel/prompts": "^0.1.24",
+        "laravel/prompts": "^0.1.24|^0.2",
```
crynobone:
```suggestion
        "laravel/prompts": "^0.1.24|^0.2.0|^0.3.0",
```

We are currently in discussion to release `0.3.0` so adding this and marking it as draft for now. I will update once we have definitive plan for week 1 of October releases.

PR #964, opened 2024-11-19
[2.x] Supports PHP 8.4 for Roadrunner and FrankenPHP

crynobone:
Release support for Roadrunner and FrankenPHP

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "laminas/laminas-diactoros": "^3.0",
         "laravel/framework": "^10.10.1|^11.0",
         "laravel/prompts": "^0.1.24|^0.2.0|^0.3.0",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -18,7 +18,7 @@
         "laminas/laminas-diactoros": "^3.0",
         "laravel/framework": "^10.10.1|^11.0",
         "laravel/prompts": "^0.1.24|^0.2.0|^0.3.0",
-        "laravel/serializable-closure": "^1.3.0",
+        "laravel/serializable-closure": "^1.3|^3.0",
```
saulens22: `laravel/serializable-closure` ^3.0 does not exist, should be ^2.0

crynobone: thanks

PR #966, opened 2024-11-21
[2.x] Supports PHP 8.4 for `swoole`

crynobone:


Cluster2a:
@taylorotwell, as far as I understand swoole v5.1.x does not support php 8.4.
Shouldn't php 8.4 only make sense using swoole v6.0.0 and upwards?

gerardnll:
Yes, only Swoole 6 is compatible with PHP 8.4.

PR #976, opened 2024-12-22
feat: add a way to define frankenphp version

pushpak1300:
Recently, due to an issue with FrankenPHP (see [issue #1276](https://github.com/dunglas/frankenphp/issues/1276)), the install command was temporarily non-functional. 

I believe there should be a way to specify the underlying FrankenPHP version in the install command. This would allow us to use a specific version if there are problems with the underlying worker servers.

Let me know if you not happy with the current implementation as I can change implementation to pass the worker server version from the command itself.


taylorotwell:

Issue #906, opened 2024-06-12
May throw BindingResolutionException when high concurrently requests in OpenSwoole


evan361425:
### Octane Version

2.4.0

### Laravel Version

11.10.0

### PHP Version

8.2.20

### What server type are you using?

Open Swoole

### Server Version

22.0.0

### Database Driver & Version

php8.2-sqlite3

### Description

When sending 10 concurrent requests into Octane server with OpenSwoole coroutine enabled, it may throw BindingResolutionException:

```shell
$php artisan octane:start --workers=2

   INFO  Server running…

  Local: http://127.0.0.1:8000 

  Press Ctrl+C to stop the server

  200    GET /google ........................................................................ 19.36 mb 189.63 ms
  200    GET /google ........................................................................ 19.37 mb 194.08 ms
  200    GET /google ........................................................................ 19.40 mb 206.60 ms
  200    GET /google ........................................................................ 19.35 mb 214.35 ms
  200    GET /google ........................................................................ 19.31 mb 221.36 ms
  200    GET /google ........................................................................ 19.26 mb 227.83 ms

   Illuminate\Contracts\Container\BindingResolutionException 

  Target [Illuminate\Contracts\Encryption\Encrypter] is not instantiable while building [Illuminate\Cookie\Middleware\EncryptCookies].

  at vendor/laravel/framework/src/Illuminate/Container/Container.php:1118
    1114▕         } else {
    1115▕             $message = "Target [$concrete] is not instantiable.";
    1116▕         }
    1117▕ 
  ➜ 1118▕         throw new BindingResolutionException($message);
    1119▕     }
    1120▕ 
    1121▕     /**
    1122▕      * Throw an exception for an unresolvable primitive.


  200    GET /google ........................................................................ 19.32 mb 240.42 ms
  200    GET /google ........................................................................ 19.29 mb 253.11 ms
  200    GET /google ........................................................................ 19.24 mb 267.67 ms
  200    GET /google ........................................................................ 19.20 mb 281.84 ms

   Illuminate\Contracts\Container\BindingResolutionException 

  Target [Illuminate\Contracts\Encryption\Encrypter] is not instantiable while building [Illuminate\Cookie\Middleware\EncryptCookies].

  at vendor/laravel/framework/src/Illuminate/Container/Container.php:1118
    1114▕         } else {
    1115▕             $message = "Target [$concrete] is not instantiable.";
    1116▕         }
    1117▕ 
[...]

Issue #765, opened 2023-10-31
Looking forward to support for Swoole coroutines


clemask:
We really look forward to Octane supporting Swoole coroutines, which will bring higher performance and solve more business scenarios. We hope the official will include this in the roadmap as soon as possible.

taylorotwell:
Feel free to work on it! 👍

clemask:
Octane is a great project that is pushing Laravel in a new direction in the realm of the internet, making Laravel more powerful. However, there seem to be some issues with the way it's being handled in certain aspects. Octane's support for Swoole has excited many developers, including myself, envisioning a strong combination of Laravel's ecosystem with Swoole's coroutine advantages.

One crucial point that is not clearly stated in the official documentation is the inability to use Swoole's coroutine. This omission is significant because developers might unknowingly configure coroutine parameters, expecting Octane to work seamlessly with Swoole's coroutine features. However, Octane does not disable the creation of Swoole services with coroutine parameters, leading to subtle issues in the production environment. Problems like network connections being used by other coroutines and issues with Monolog log cycling detection can arise. In the end, developers are forced to disable coroutines as a quick solution, which is a severe incident.

I hope the official team promptly considers adding coroutine support to their roadmap. It's a fantastic feature that not only enhances Laravel but is also crucial for the broader PHP ecosystem.

driesvints:
> I hope the official team promptly considers adding coroutine support to their roadmap.

This isn't a trivial thing to do. Right now we don't have the time to work on this ourselves but we'd love to see PR's kickstarting this.

driesvints:
Welcoming prs 👍

binaryfire:
@driesvints I've been active in the Hyperf community lately. I'll put the word out and see if any Swoole gurus can help.

huangdijia:
If you want octane to support coroutine, you must first make database, redis, etc. support connection pooling.

binaryfire:
I guess the problem is Laravel supports a lot of DB and cache backends and it’d be hard to implement pooling for them all…

@taylorotwell @driesvints @nunomaduro Would the core team be interested in having pooling for a select few drivers? Maybe just MySQL and Redis to start. It’d mean coroutines could only be enabled when using those drivers, but Hyperf can handle 100k req/second. Even if coroutines ‘only’ bumped Laravel’s concurrency to 10k-20k, that’d be nuts and create some serious buzz. That’s faster than Node…

[...]

PR #977, opened 2024-12-23
[WIP] Enable WebSockets via Swoole

kingIZZZY:
## [WIP] Laravel Octane Swoole WebSockets

Already using Laravel Octane via Swoole?

Take full advantage of Swoole built-in WebSockets features to serve Laravel powered WebSockets!

Configure `octane.swoole.enableWebSockets` to `true` to enable

### DIFF
--- a/src/Worker.php
+++ b/src/Worker.php
@@ -12,6 +12,8 @@ use Laravel\Octane\Events\TaskReceived;
 use Laravel\Octane\Events\TaskTerminated;
 use Laravel\Octane\Events\TickReceived;
 use Laravel\Octane\Events\TickTerminated;
+use Laravel\Octane\Events\WebSocketDisconnectReceived;
+use Laravel\Octane\Events\WebSocketMessageReceived;
 use Laravel\Octane\Events\WorkerErrorOccurred;
 use Laravel\Octane\Events\WorkerStarting;
 use Laravel\Octane\Events\WorkerStopping;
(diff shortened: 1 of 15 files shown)


kingIZZZY:
To access the Swoole server object from laravel code - here is one way how:
```php
app('Laravel\Octane\Swoole\WorkerState')->server
```
For example to push WebSocket messages to clients:
```php
app('Laravel\Octane\Swoole\WorkerState')->server->push($clientSocketFD, $message);
```

Is this the 1) only way? 2) correct way? If anyone knows - please share 🙏

src/Worker.php
```diff
@@ -241,4 +243,44 @@ public function terminate(): void
     {
         $this->dispatchEvent($this->app, new WorkerStopping($this->app));
     }
+
+    /**
+     * Handle Incoming WebSocket Messages
+     */
+    public function handleWebSocketMessage(\Swoole\WebSocket\Server $server, \Swoole\WebSocket\Frame $frame)
+    {
+        \Laravel\Octane\CurrentApplication::set($sandbox = clone $this->app);
```
kingIZZZY, line 252: Pretty much copied from `handleTick`/`handleTask`, does it make sense? Should it be doing these same steps for incoming WebSocket messages/disconnect?

taylorotwell:
Would need to be Echo compatible IMO.

kingIZZZY:
Could we keep the PR open for a bit longer to 1) maybe hear some more opinions 2) maybe attract some more attention of others better well versed with Echo than myself who can take up contributing on that aspect?

Or is there an evident impossibility of compatibility with Echo which dooms this whole approach/effort that I am missing?

goodevilgenius:
Personally, I think this looks like a really good start to me.

Only suggestion I have is that I don't think the union type for `Server|WServer` is necessary. Last time I checked, `Swoole\WebSocket\Server` extends `Swoole\Http\Server`. So, you can just type check the `Http` server and that covers both.

Making it work with echo would be the logical next step, though.

kingIZZZY:
Oh no 😭 I think I might abandon this effort for now
It seems this feature is prone to [throwing this exception](https://github.com/laravel/octane/issues/906) possibly among others:
- BindingResolutionException: Target class [events] does not exist.

Because: Laravel does not support Swoole Coroutines
- https://github.com/laravel/octane/issues/765 specifically [this comment](https://github.com/laravel/octane/issues/765#issuecomment-2417130074)

Seems it might be a deal breaker for now, oh well - it's back to [Hyperf](https://github.com/hyperf/hyperf) for me...
Anyone who'd like can also check out the [Laravel-Hyperf](https://laravel-hyperf.com/) project (see the above linked issue #765 and specific comment)

izdrail:
It's a nice project Hyperf. 
I like roardrunner better i don't know why

PR #982, opened 2025-01-06
Feat: add support for Swoole task_enable_coroutine & task_use_object/task_object option

nguyenhothanhtam0709:
This PR modify on task callback of Swoole server to support options `task_enable_coroutine` & `task_use_object` set to `true`. As you can see in [Swoole document](https://wiki.swoole.com/en/#/server/setting?id=task_enable_coroutine). When the option `task_enable_coroutine` is enabled, php code can directly use Swoole coroutine api in onTask callback, as well as in Octane::concurrently like bellow code:
```php
Route::get(('/get-post'), function () {
    [$msg] = Octane::concurrently([
        function () {
            $msg = Post::all();
            Log::info('TASK DONE!');
            return $msg;
        },
        function () {
            Coroutine::sleep(1);
        }
    ]);

    return [
        "message" => "ok",
        "data" => $msg
    ];
});
```

The option `task_use_object`/`task_object` allows to use the same callback like option `task_enable_coroutine`

### DIFF
--- a/bin/swoole-server
+++ b/bin/swoole-server
@@ -9,6 +9,7 @@ use Laravel\Octane\Swoole\ServerStateFile;
 use Laravel\Octane\Swoole\SwooleExtension;
 use Laravel\Octane\Swoole\WorkerState;
 use Swoole\Http\Server;
+use Swoole\Server\Task;
 use Swoole\Timer;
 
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #983, opened 2025-01-11
Add manifest.json to watch default list

carferdas:
Currently many of the projects work with Vite, currently if you build the manifest file it is modified and not watch that the previous files were deleted and new ones exist in the `manifest.json`.


taylorotwell:
Hmm, I don't quite follow.

PR #984, opened 2025-01-13
commands/start-swoole: allow customization of log-level

rochacon:
The current start command for Swoole mode does not allow customization of the log level via the options array. 

This change add the same logic as we have for FrankenPHP and RoadRunner, giving precedence to the `log-level` option and falling back to the environment name as previous behavior.

### DIFF
--- a/src/Commands/StartSwooleCommand.php
+++ b/src/Commands/StartSwooleCommand.php
@@ -28,7 +28,8 @@ class StartSwooleCommand extends Command implements SignalableCommandInterface
                     {--task-workers=auto : The number of task workers that should be available to handle tasks}
                     {--max-requests=500 : The number of requests to process before reloading the server}
                     {--watch : Automatically reload the server when the application is modified}
-                    {--poll : Use file system polling while watching in order to watch files over a network}';
+                    {--poll : Use file system polling while watching in order to watch files over a network}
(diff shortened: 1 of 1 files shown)


taylorotwell:
Wouldn't it need to be added as an available option and passed to the command?

rochacon:
I believe you're right @taylorotwell, I copied the check from [`src/Commands/StartFrankenPhpCommand.php`](https://github.com/laravel/octane/blob/28a313abab7b8cfdbac868cc58dce535e5b437af/src/Commands/StartFrankenPhpCommand.php#L102) and forgot to validate the flag. It should be correct now.

taylorotwell:

Cluster2a:
> Thanks for your pull request to Laravel!
> 
> Unfortunately, I'm going to delay merging this code for now. To preserve our ability to adequately maintain the framework, we need to be very careful regarding the amount of code we include.
> 
> If applicable, please consider releasing your code as a package so that the community can still take advantage of your contributions!

Did you even take a look at the code?

MaddoxCH:
@taylorotwell The code is **literally** the same one used in StartRoadrunnerCommand.php, and you **already** accepted the same merge request for roadruner a while ago.
https://github.com/laravel/octane/commit/82a5b680e76e96b560d695be417d0176057b71e6

PR #994, opened 2025-02-10
Revert "Clear resolved Vite instance (via Facade) to avoid memory-leak"

taylorotwell:
Reverts laravel/octane#993


markieo1:
Is there a reason for the revert? We are currently seeing the following issue:
```
Unable to locate file in Vite manifest: src/app.js
```
And are wondering if this has something to do with this?

crynobone:
The changes was never released on Octane

PR #1008, opened 2025-03-08
Set version=3 for RoadRunner's reload

apih:
Since RoadRunner v2024.3.0 and above, running `octane:reload` command will result in an error, where the error message is `Cannot reload RoadRunner: rr configuration file should contain a version e.g: version: 3`. Setting the version when calling RoadRunner binary in `src/RoadRunner/ServerProcessInspector.php` file seems to fix it.

image

### DIFF
--- a/src/RoadRunner/ServerProcessInspector.php
+++ b/src/RoadRunner/ServerProcessInspector.php
@@ -47,7 +47,7 @@ class ServerProcessInspector implements ServerProcessInspectorContract
         tap($this->processFactory->createProcess([
             $this->findRoadRunnerBinary(),
             'reset',
-            '-o', "version=3",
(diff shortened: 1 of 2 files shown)


taylorotwell:
It seems like this might break people not using that version?

taylorotwell:
Tests failing.

apih:
> It seems like this might break people not using that version?

But currently, the version is already set to 3 in [`src/Commands/StartRoadRunnerCommand.php#84`](https://github.com/laravel/octane/blob/9ce570a20f6c36c05d0b485f40fc48d89c96655d/src/Commands/StartRoadRunnerCommand.php#L84). So the RoadRunner process will be started in version 3.

Issue #1004, opened 2025-03-02
No ability to deploy without downtime


mikkpokk:
### Octane Version

2.8.1

### Laravel Version

10.48.28

### PHP Version

8.3.16

### What server type are you using?

Swoole

### Server Version

6.0.1

### Description

I'm using Envoyer to deploy a Laravel Octane application, taking advantage of its zero-downtime deployment features.

However, Octane currently does not support zero-downtime deployment because it does not follow symlink directories. It always remains in the actual directory (instead of the symlinked one) where the Octane server was initially started. When the outdated release directory is deleted, Octane continues running in that location, causing errors on every request and resulting in 500 response codes on the live application.

Error thrown:
```
Warning: require(/var/www/domain.com/releases/202502010200023/vendor/laravel/octane/bin/bootstrap.php): Failed to open stream: No such file or directory in /var/www/domain.com/releases/202502010200023/vendor/laravel/octane/bin/swoole-server on line 18
Fatal error: Uncaught Error: Failed opening required '/var/www/domain.com/releases/202502010200023/vendor/laravel/octane/bin/bootstrap.php' (include_path='.:/usr/bin/php@8.3/8.3.16/share/php@8.3/pear') in /var/www/domain.com/releases/202502010200023/vendor/laravel/octane/bin/swoole-server:18
Stack trace:
#0 /var/www/domain.com/releases/202502010200023/vendor/laravel/octane/bin/swoole-server(95): {closure}(Array)
#1 [internal function]: {closure}(Object(Swoole\Http\Server), 0)
#2 /var/www/domain.com/releases/202502010200023/vendor/laravel/octane/bin/swoole-server(170): Swoole\Server->start()
#3 {main}
  thrown in /var/www/domain.com/releases/202502010200023/vendor/laravel/octane/bin/swoole-server on line 18
#1 [internal function]: {closure}(Object(Swoole\Http\Server), 1)
#1 [internal function]: {closure}(Object(Swoole\Http\Server), 3)
#1 [internal function]: {closure}(Object(Swoole\Http\Server), 2)
#1 [internal function]: {closure}(Object(Swoole\Http\Server), 4)
#1 [internal function]: {closure}(Object(Swoole\Http\Server), 5)
```

### Steps To Reproduce

Use any 0-downtime deployment or test it manually using the following instructions:

1. Use the Swoole driver as an example.
2. `cd` one directory up from the project's base path.
[...]

PR #1009, opened 2025-03-12
Swoole: Follow symlinks to enable zero-downtime deployment

mikkpokk:
Issue is addressed in #1004. This PR fixes the addressed issue.  

What does this PR do?  
- If PHP `SCRIPT_NAME` contains the base directory and it differs from `basePath`, it replaces realpath paths with symlinked paths.  
- Detects if the server is inside a symlinked directory and replaces realpaths with symlinked paths.  
- Forwards the correct directory path to `swoole-server` using the environment parameter `APP_RELEASE_BIN_DIR`.  
- `swoole-server` uses `APP_RELEASE_BIN_DIR` instead of the PHP realpath `__DIR__`.  
- `swoole-server` listens to the Swoole event `beforereload`. Whenever the app is reloaded, it clears the PHP directory caches using `clearstatcache(true)`.


taylorotwell:
Hello! Are you able to update this PR to follow Laravel's code style conventions?

mikkpokk:
@taylorotwell Hello! Done. Added the necessary PHPDoc and verified the code with Laravel Pint using `vendor/bin/pint`.

I've no idea why the Roadrunner tests on the cloud are failing randomly. They are not related at all. Most likely caused by #1008

src/Commands/StartSwooleCommand.php
```diff
@@ -138,6 +180,54 @@ protected function defaultServerOptions(SwooleExtension $extension)
...
+                $this->pwd = "$cwd/$working_dir";
+            } else {
+                $this->pwd = $working_dir;
+            }
+        } elseif ($working_dir === '.') {
+            // This part comes into action only if the server changes to
+            // the base path directory before running the Artisan command.
+            // eg. cd /www/current && php artisan octane:start
+            if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') {
+                $this->pwd = trim(shell_exec('cd')); // For Windows
```
osbre, line 222: According to other PRs:

> Octane does not support Windows.

Therefore, it would make no sense to include any partial Windows support.

src/Commands/StartSwooleCommand.php
```diff
@@ -44,6 +44,47 @@ class StartSwooleCommand extends Command implements SignalableCommandInterface
...
+    /**
+     * The reliable working directory path of the process.
+     * Respects the symlink path and does not convert to the real path.
+     */
+    protected string $pwd;
+
+    public function __construct()
+    {
+        $this->dir = __DIR__;
+        $this->real_cwd = base_path();
```
osbre, line 71: Why would you set it to one value, only to re-assign it on the next line in `getPwd`?

Looking at the name, `getPwd` should ***return***/get something, not set it.

Also, If applicable, it's best to avoid unnecessary nesting levels by embracing [early-returns](https://freek.dev/1593-writing-readable-php-decrease-indentation-by-returning-early).

src/Commands/StartSwooleCommand.php
mikkpokk: Using Roadrunner with Octane requires the pcntl extension, which is unavailable on Windows. This isn't the case with Swoole, however.

src/Commands/StartSwooleCommand.php
mikkpokk: `$this->real_cwd` is not the same as `$this->pwd`.

Moreover, we cannot talk about early returns in the constructor function at all. 

EDIT: You cannot and should not return anything in the constructor.

sy-records:
> I've no idea why the Roadrunner tests on the cloud are failing randomly. They are not related at all. Most likely caused by #1008

Yep, I fixed it. #1023

taylorotwell:

Muffinman:
I'm hoping this can be looked at again in the future.

Reloading without downtime can be critical requirement for high traffic environments, which is exactly where octane is most heavily used.

Presumably Laravel themselves came across the issue and had to solve it when adding Octane to Laravel Cloud?

PR #1049, opened 2025-08-04
Make Swoole package_max_length configurable via config and env

mohammed-samir01:
This pull request adds the ability to configure the package_max_length option for Swoole directly from the config/octane.php configuration file, or by using the OCTANE_PACKAGE_MAX_LENGTH environment variable.

Motivation
Previously, the maximum allowed size for incoming requests in Swoole was not easily configurable. This change provides users with more flexibility to control upload limits based on their application needs, especially for handling large file uploads.

How it works
By default, the value is set to 50 MB. To customize, simply set the environment variable in your .env file


taylorotwell:

PR #1058, opened 2025-09-21
Refactor error handling and environment variable access

7amoood:
Access ENV vars using $_ENV require more additional settings in  php.ini to add "E" to variables_order so we can using getenv instead


7amoood:
@taylorotwell I have extended start swoole server to change some env vars without any effect, I realized that I have to open flag E in php.ini or using getenv instead of $_ENV

PR #1063, opened 2025-10-06
FrankenPHP: Add env option to disable Caddy access logs

athifsha:
This PR adds support for disabling Caddy access logs when running Laravel Octane with FrankenPHP.

Problem:
Caddy logs every request by default, creating unnecessary noise in production logs where centralized logging (Grafana, NewRelic, etc.) is already used.

Solution:
Added an environment variable OCTANE_FRANKENPHP_DISABLE_ACCESS_LOGS that, when set to true, disables the access logs by commenting out the log {} block in the generated Caddyfile.

Behavior:

Default: logs enabled (backward-compatible).

When OCTANE_FRANKENPHP_DISABLE_ACCESS_LOGS=true, access logs are disabled.

Testing:
Verified locally with Octane + FrankenPHP; access logs were correctly suppressed when the env was set.

Changelog:
Added OCTANE_FRANKENPHP_DISABLE_ACCESS_LOGS environment variable to disable Caddy access logs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased](https://github.com/laravel/octane/compare/v2.12.3...2.x)
 
+### Added
+
+* FrankenPHP: Add env option `OCTANE_FRANKENPHP_DISABLE_ACCESS_LOGS` to disable Caddy access logs
+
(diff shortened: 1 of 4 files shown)


taylorotwell:
Just change it in your Caddyfile.

athifsha:
Got it 👍 makes sense — I’ll move the access log disable logic to our own Caddyfile instead. Thanks for the clarification!

Issue #1064, opened 2025-10-09
Octane with Swoole - Defaultvalues in Signature prevent uses config values


ml4u:
### Octane Version

2.11.0

### Laravel Version

12.33.0

### PHP Version

8.4.10

### What server type are you using?

Swoole

### Server Version

-

### Database Driver & Version

-

### Description

Hi,
i found a bug in StartCommand Class of Octane.

```php
    protected function startSwooleServer()
    {
        return $this->call('octane:swoole', [
            '--host' => $this->getHost(),
            '--port' => $this->getPort(),
            '--workers' => $this->option('workers') ?: config('octane.workers', 'auto'),
            '--task-workers' => $this->option('task-workers') ?: config('octane.task_workers', 'auto'),
            '--max-requests' => $this->option('max-requests') ?: config('octane.max_requests', 500),
            '--watch' => $this->option('watch'),
            '--poll' => $this->option('poll'),
        ]);
    }
```

Because of defaultvalues in $signature, will the option() call never be null and skips the config values.

### Steps To Reproduce

You can set in config/octane.php 
- workers: 100

dump what you call in startSwooleServer(). Workers will be auto.
Only overrideable with explicit argument like
$ php artisan octane:start --server=swoole --workers=100

PR #1070, opened 2025-10-22
Fix: Octane ignores workers, task_workers, max_requests values from config

pr-man:
**Problem**

When running `php artisan octane:start` without passing the `--workers`, `--task-workers`, `--max-requests` options, Octane does not use the values from the configuration (`octane.workers`, `octane.task_workers`, `octane.max_requests`). This happens because the options always return non null values, so the code never falls back to the config values.

**Solution**

Remove the default values from the `--workers`, `--task-workers`, `--max-requests` options. The default value is already provided in `config()` fallback.

```
'--workers' => $this->option('workers') ?: config('octane.workers', 'auto'),
'--task-workers' => $this->option('task-workers') ?: config('octane.task_workers', 'auto'),
'--max-requests' => $this->option('max-requests') ?: config('octane.max_requests', 500)
```

Issue: #1064
Follow-up to the pull request: #1028


src/Commands/StartCommand.php
```diff
@@ -22,8 +22,8 @@ class StartCommand extends Command implements SignalableCommandInterface
                     {--admin-port= : The port the admin server should be available on [FrankenPHP only]}
                     {--rpc-host= : The RPC IP address the server should bind to}
                     {--rpc-port= : The RPC port the server should be available on}
-                    {--workers=auto : The number of workers that should be available to handle requests}
-                    {--task-workers=auto : The number of task workers that should be available to handle tasks}
+                    {--workers= : The number of workers that should be available to handle requests}
+                    {--task-workers= : The number of task workers that should be available to handle tasks}
                     {--max-requests=500 : The number of requests to process before reloading the server}
```
sy-records:
```suggestion
                    {--max-requests=: The number of requests to process before reloading the server}
```

sy-records (commented):
The same issue.

taylorotwell:
Since we don't even document this config based approach I would rather just remove it. Just pass the workers you want via CLI.

pr-man:
@taylorotwell Since this decision has been made, it would be nice to include config support in future branches. Laravel Octane is still IO-blocking, and sometimes it’s necessary to keep more workers than there are CPUs. Imagine if you have many projects or, for example, different pod configurations in Kubernetes — it would be convenient to use constructions like this without having to adjust everything manually each time:

```
'server' => env('OCTANE_SERVER', 'swoole'),
'workers' => env('OCTANE_WORKERS', swoole_cpu_num() * 4),
'task_workers' => 1,
```

PR #1088, opened 2026-01-08
[2.x] Add GiveNewApplicationInstanceToEventDispatcher listener

stevethomas:
## Summary

This PR adds a `GiveNewApplicationInstanceToEventDispatcher` listener that updates the EventDispatcher's queue resolver to use the sandbox application instance.

## Problem

The EventDispatcher's `queueResolver` closure is set during boot in Laravel's `EventServiceProvider` and captures the original application instance. When Octane creates a sandbox for each request and queue configuration is modified during that request, queued event listeners still dispatch to the original app's queue manager, ignoring the sandbox's configuration changes.

## Solution

Following the established pattern of `GiveNewApplicationInstanceToQueueManager` and other similar listeners, this PR adds a listener that updates the EventDispatcher's queue resolver to use the sandbox's queue manager for each request.

## Benefit to End Users

Applications that modify queue configuration during a request (e.g., dynamically switching queue connections based on tenant or request context) will now have their queued event listeners respect those configuration changes when running under Octane.

## Why This Doesn't Break Existing Features

- The listener follows the exact same pattern as existing listeners (`GiveNewApplicationInstanceToQueueManager`, `GiveNewApplicationInstanceToMailManager`, etc.)
- It only executes if the `events` service has been resolved
- It simply updates the queue resolver callable, which is the same operation Laravel performs during boot
- The listener is registered after `GiveNewApplicationInstanceToQueueManager` to ensure the queue manager is ready

## Files Changed

- `src/Listeners/GiveNewApplicationInstanceToEventDispatcher.php` - New listener
- `src/Concerns/ProvidesDefaultConfigurationOptions.php` - Register the listener
- `tests/EventDispatcherStateTest.php` - Test verifying the queue resolver uses the sandbox

### DIFF
--- a/src/Concerns/ProvidesDefaultConfigurationOptions.php
+++ b/src/Concerns/ProvidesDefaultConfigurationOptions.php
@@ -42,6 +42,7 @@ trait ProvidesDefaultConfigurationOptions
             \Laravel\Octane\Listeners\GiveNewApplicationInstanceToCacheManager::class,
             \Laravel\Octane\Listeners\GiveNewApplicationInstanceToSessionManager::class,
             \Laravel\Octane\Listeners\GiveNewApplicationInstanceToQueueManager::class,
+            \Laravel\Octane\Listeners\GiveNewApplicationInstanceToEventDispatcher::class,
             \Laravel\Octane\Listeners\GiveNewApplicationInstanceToRouter::class,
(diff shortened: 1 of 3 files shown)


taylorotwell:
Fixed in framework here: https://github.com/laravel/framework/commit/bdc94cee3cd0a4b08df55858ee7aacc5c731f444

Thanks

PR #1093, opened 2026-02-06
[2.x] Supports Laravel 13

crynobone:
* [x] `livewire/livewire`
* [x] `inertiajs/inertia-laravel`

PR #1103, opened 2026-03-18
Support PHPUnit 13 and fix Laravel Passport compatibility conflicts

Nationalcat:
## What Changed

1. Allowed PHPUnit 13 in require-dev by updating the version constraint:
2. Removed the Composer conflict that blocked PHPUnit versions above 12.
3. Removed all `@doesNotPerformAssertions` / `#[DoesNotPerformAssertions]` usages from tests.
4. Added explicit assertions in tests that previously relied on those markers.
5. Fixed WorkerTest output-buffer handling to avoid closing PHPUnit-managed buffers.
6. Updated FlushStrCacheTest to verify cache flushing between requests in a stable way.
7.  Allowed `symfony/psr-http-message-bridge ^8.0` to resolve Laravel Passport transitive dependency conflicts.

## Why
- Previously, the project did not have an explicit PHPUnit 13 verification path in CI, and Composer settings prevented installing PHPUnit > 12.
- This PR ensures PHPUnit 13 compatibility is both supported and continuously validated.

PR #1104, opened 2026-03-18
[2.x] Remove conflict requirement for phpunit/phpunit

joelbutcher:
Address the comments in #1093 and removes the PHPUnit conflict rule, which directly conflicts with the Laravel Skeleton, meaning Octane doesn't truly support L13 without first requiring a downgrade in PHPUnit version (which seems iffy)

I don't **_actually_** know why this was added, there's no mention of it in the original PR.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,6 @@
         "bin/swoole-server"
     ],
     "conflict": {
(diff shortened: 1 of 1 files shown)


eshimischi:
https://github.com/laravel/octane/pull/1103 here already

PR #1112, opened 2026-03-23
Move worker/request config resolution to server-specific commands

kohlerdominik:
Previously, `octane.workers`, `octane.task_workers`, and `octane.max_requests` were only read from config inside `StartCommand`. This meant that running a server directly via `octane:swoole`, `octane:roadrunner`, or `octane:frankenphp` would ignore these config values entirely and always fall back to hardcoded defaults.

This is problematic in containerized environments, where you want to control worker count via environment variables (e.g. `OCTANE_WORKERS=4`) without hardcoding CLI flags to the start command.

### Changes

- Move config resolution into each server-specific command (`StartSwooleCommand`, `StartRoadRunnerCommand`, `StartFrankenPhpCommand`), so it works regardless of which command is used to start the server
- Add `workers`, `task_workers`, and `max_requests` to `config/octane.php` with `env()` support (`OCTANE_WORKERS`, `OCTANE_TASK_WORKERS`, `OCTANE_MAX_REQUESTS`)
- CLI flags still take precedence over config values

### Supported now

```bash
# Works for all server commands
OCTANE_WORKERS=4 php artisan octane:frankenphp
OCTANE_WORKERS=4 php artisan octane:roadrunner
OCTANE_WORKERS=4 php artisan octane:swoole

# CLI flag still overrides config
php artisan octane:frankenphp --workers=8
```

If this gets merged, also look at the corresponding docs update: https://github.com/laravel/docs/pull/11126


taylorotwell:
I find this one a bit confusing. Can we just make the start command work how you want?

kohlerdominik:
I found it very confusing, that Claude Code suggested to configure `octane.workers`, `octane.task_workers`, and `octane.max_requests` but that didn't work with the suggested command from the [docs ](https://laravel.com/docs/13.x/octane#frankenphp-via-docker):
```php
```

So I tried to make it more consistent overall, by allowing the config variables from all public entrypoints instead of just one. Or are the `octane:frankenphp`, `octane:swoole` and `octane:roadrunner` commands not meant to be used?

Issue #1115, opened 2026-04-12
File watcher glob workaround doesn't activate for chokidar v5


mikerockett:
### Octane Version

2.17.1

### Laravel Version

13

### PHP Version

8.5

### What server type are you using?

Swoole

### Server Version

6.0

### Description

The fix for #962 (glob support removed in chokidar v4) only checks `version.startsWith('4.')` in `bin/file-watcher.cjs` line 8. Chokidar v5 also removed glob support, so the path extraction and `ignored` filter never activate when chokidar v5 is installed.

This means wildcard watch paths like `app-modules/**/*.php` or `config/**/*.php` are passed as raw globs to chokidar v5, which silently ignores them. Only plain directory paths (e.g. `app`, `routes`) work.

The version check should use a semver comparison (`>= 4`) rather than a string prefix check.

### Steps To Reproduce

1. Install chokidar v5 (`npm install chokidar@^5.0.0`)
2. Keep the default `config/octane.php` watch paths (which include `config/**/*.php`, `database/**/*.php`, etc.)
3. Run `octane:start --watch`
4. Modify a file in `config/` - no reload
5. Modify a file in `app/` (no glob) - reload works
6. Change `config/**/*.php` to `config` in the watch array - reload now works for config changes

### Suggested Fix

In `bin/file-watcher.cjs`, change:

```js
const chokidarVersion4 = require(chokidarPackagePath).version.startsWith('4.');
```

To something like:

```js
const chokidarMajor = parseInt(require(chokidarPackagePath).version.split('.')[0], 10);
const chokidarVersion4 = chokidarMajor >= 4;
```

PR #1117, opened 2026-04-16
[2.x] Restore Glob Support For Chokidar v5

eagostini:
### Description

This pull request restores and future-proofs the file-watching mechanism for users on modern versions of Chokidar (v4 and the recently released v5). It should handle #1115.

While Chokidar removed native glob support starting in version 4, the previous implementation of the watcher script had a version-specific check that hardcoded a search for version `4.x`. When Chokidar v5 was released, this check failed. This caused the watcher to attempt to pass glob patterns directly to a library that no longer understands them, resulting in a `TypeError`.

This commit improves the version detection to apply a manual glob polyfill for all versions greater than 3.

### Benefits to End Users

* **Compatibility:** Restores file-watching functionality for developers on the latest Node environments.
* **Future-proofing:** Using a generic greater-than-3 version check prevents this exact regression from happening on future Chokidar major releases.
* **Efficiency:** The custom path parsing logic is completely bypassed if the user is running Chokidar v3, avoiding any unnecessary regex execution.

### Why This Does Not Break Existing Features

The manual glob polyfill is strictly conditional. Users on Chokidar v3 will continue to use the native glob engine without any changes to their execution path. The script also introduces zero new dependencies to the repository.

### Tests

As this is a standalone Node script, I manually verified the fix across multiple environments within a Docker container to ensure no regressions:

* **Verified on Chokidar v5.0.0:** Confirmed it resolves the `TypeError` and correctly filters files via the new polyfill.
* **Verified on Chokidar v4.0.0:** Confirmed the polyfill correctly handles the missing glob support.
* **Verified on Chokidar v3.6.0:** Confirmed the script skips the polyfill entirely and uses the native glob engine as expected.

**Verified Scenarios:**
I successfully tested all path definitions that ship by default in the `config/octane.php` file, including:
1. Standard files (e.g., `composer.lock`)
2. Simple directories (e.g., `app`)
3. Wildcard patterns (e.g., `config/**/*.php`)
4. Blade templates (e.g., `resources/**/*.blade.php`)

### DIFF
--- a/bin/file-watcher.cjs
+++ b/bin/file-watcher.cjs
@@ -5,7 +5,7 @@ const poll = process.argv[3] ? true : false;
 
 // Chokidar removed support for glob in version 4...
 const chokidarPackagePath = require.resolve('chokidar').replace('index.js', 'package.json');
-const chokidarVersion4 = require(chokidarPackagePath).version.startsWith('4.');
+const chokidarVersion4 = parseInt(require(chokidarPackagePath).version.split('.')[0], 10) >= 4;
 
 const extractWildcardExtension = (path) => {
     // Match patterns like *.php, **/*.blade.php


taylorotwell:
Since this seems agent generated, here is my agent's response:

I found a blocking regression in this PR.

- bin/file-watcher.cjs:29 extracts extension with path.match(/^.+?(?<=\.)(?<extension>.+)/), which matches the first dot anywhere in the absolute path, not just wildcard extensions.
- Watch paths are absolute (base_path(...) in src/Commands/Concerns/InteractsWithServers.php:91), so a project path like /tmp/my.app produces bad extensions (app/app, app/config/**/*.php, etc.).
- Then options.ignored at bin/file-watcher.cjs:43 treats those as required suffixes and ignores real file changes (notably directory watches like app/config), so reloads stop working on dotted project paths.
- Repro example: project in /tmp/my.app with watch entry app -> any /tmp/my.app/app/*.php file change is ignored.
Suggested fix:
- Only compute extension for wildcard patterns (like the previous extractWildcardExtension logic), instead of any path containing a dot.
- Keeping the new major-version check (> 3) is good; the regression is in extension parsing scope, not the version gating.
If you want, I can also draft a ready-to-post PR review comment for GitHub with this repro and a patch suggestion.

eagostini:
Hi @taylorotwell, thanks for reviewing. Just noticed you sent a patch through earlier to sort this issue. Do you feel the performance improvements are still relevant? If so, I can put through a new PR for that.

About the PR description... read it again this morning and yeah, it's too verbose 😅

Cheers 🥂

PR #1124, opened 2026-05-16
feat: App instance safety

AlliBalliBaba:
This PR is a simplified version of #980 containing only the safer concept around the app container.

Currently, the app container is cloned on each request and the new instance is explicitly passed through event listeners. The disadvantage of this approach is that it's easy to leak memory and state between requests by keeping a stale reference to the old app container somewhere in the code.

This approach instead only clones the application instance once when it first is captured. Afterwards it resets all properties on the container instance directly through an inheritance trick. This makes Octane safer by preventing any leaks of the application instance and faster by not requiring as many `resetApplicationInstance` event listeners.


taylorotwell:

PR #1126, opened 2026-05-22
Skip full compiled route iteration in router listener

faytekin:
Fixes #1125.

#1121 fixed a real warmed-routes container leak, but the trailing `foreach ($routes as $route)` loop forces full hydration of every compiled route on every request, which defeats the lazy design and causes a regression in route-heavy apps (+20 ms p50 / +143 ms p95 for ~1200 routes in production).

This PR keeps the warmed-routes fix (the `nameCache` iteration) but skips the full collection iteration. Dev-mode `RouteCollection` behavior is preserved as before #1121.

All existing tests pass.


mertasan:
same issue here on 2.17.3 with a route-heavy app, confirming the regression. hoping to see this merged soon.

PR #1144, opened 2026-06-08
Fix FrankenPHP streamed generator responses

tecbeast42:
Fixes laravel/octane#1142.

### What changed

This updates the FrankenPHP client to handle `StreamedResponse` callbacks that return a generator or string directly. Laravel Framework's `response()->stream()` intentionally installs generator callbacks directly while `LARAVEL_OCTANE` is set. Under FrankenPHP, delegating those responses to Symfony's `send()` sends headers but does not write the returned generator chunks, producing a `200 OK` response with an empty body.

The new path is limited to `StreamedResponse` callbacks that can be safely detected as returning `Generator` or `string`. Normal responses and traditional echo-based `StreamedResponse` callbacks continue through the existing `$response->send()` behavior.

### End-user benefit

This fixes streamed responses such as SSE/event streams under Octane with FrankenPHP. In practice, generator-based streams can currently appear to succeed while returning an empty body, which breaks features like chat/event streaming responses.

### Compatibility

This mirrors the existing RoadRunner behavior, which already detects generator/string streamed callbacks and sends them through RoadRunner's streaming response path. Echo-based streaming remains untouched because callbacks without a supported return type still fall back to Symfony's normal `send()` handling.

### Tests

Added coverage for:

- Direct FrankenPHP streamed generator callbacks.
- Direct FrankenPHP streamed string callbacks.
- Worker-level streamed responses created by Laravel's response factory while `LARAVEL_OCTANE` is set.

### DIFF
--- a/src/FrankenPhp/FrankenPhpClient.php
+++ b/src/FrankenPhp/FrankenPhpClient.php
@@ -2,13 +2,16 @@
 
 namespace Laravel\Octane\FrankenPhp;
 
+use Generator;
 use Illuminate\Foundation\Application;
 use Illuminate\Http\Request;
 use Laravel\Octane\Contracts\Client;
 use Laravel\Octane\Octane;
 use Laravel\Octane\OctaneResponse;
 use Laravel\Octane\RequestContext;
+use ReflectionFunction;
 use Symfony\Component\HttpFoundation\Response;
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #1149, opened 2026-07-08
fix: avoid Swoole write() when coroutines are disabled

L3o-pold:
## Problem

When `enable_coroutine` is `false` (the default set in `StartSwooleCommand`), Swoole's `Response::write()` fails with:

```
Swoole\Error: API must be called in the coroutine
```

This crash occurs for **any response larger than `chunkSize` (1MB)**, since `SwooleClient::sendResponseContent()` falls back to `write()` for chunked sending at line 266.

### Stack trace (production)

```
PHP Fatal error: Uncaught Swoole\Error: API must be called in the coroutine
  in vendor/laravel/octane/src/Swoole/SwooleClient.php:266
#0 Swoole\Http\Response->write('z98s.ftp.infoma...')
#1 SwooleClient->sendResponseContent(...)
#2 SwooleClient->respond(...)
#3 Worker->handle(...)
```

## Root Cause

`StartSwooleCommand::defaultServerOptions()` sets:

```php
'enable_coroutine' => false,
'send_yield'       => true,
```

With coroutines disabled, the request handler runs in a synchronous context. However, `sendResponseContent()` calls `\$swooleResponse->write()` in **three** places, all of which require an active coroutine:

| Line | Context | When triggered |
|------|---------|----------------|
| 230 | `outputBuffer` | Any echoed output during request handling |
| 236 | `StreamedResponse` | All streamed responses |
| 266 | Large response chunking | Responses > 1MB |

## Fix

Introduces a `$canStream` flag based on the `enable_coroutine` config value. When coroutines are disabled, all `write()` calls are replaced with `end()`:

1. **outputBuffer**: prepended to response content, sent via `end()`
2. **StreamedResponse**: buffered with `ob_start()`/`ob_get_clean()`, sent via `end()`
3. **Large response chunking**: full content sent via `end()` instead of chunks

### Trade-off

Without coroutines, large responses are sent in a single `end()` call rather than streamed chunks. This is acceptable because:
- Responses ≤ 1MB already use `end()` exclusively
- The worker is busy until the full response is sent either way
- Crashing is worse than blocking

## Verification

Reproduced with a standalone Swoole server testing three scenarios (60MB payload):

| # | Scenario | Coroutines | Fix | Result |
|---|----------|------------|-----|--------|
| 1 | Old logic (the bug) | off | no | **FAIL** — 327KB received, server crashed |
| 2 | Fixed logic (the fix) | off | yes | **PASS** — 60MB received |
| 3 | Old logic (workaround) | on | no | **PASS** — 60MB received |

All three scenarios matched expectations.

### DIFF
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -226,6 +226,8 @@ class SwooleClient implements Client, ServesStaticFiles
             return;
         }
 
+        $canStream = config('octane.swoole.options.enable_coroutine', false);
+
         if ($octaneResponse->outputBuffer) {
             $swooleResponse->write($octaneResponse->outputBuffer);
         }
@@ -256,7 +258,7 @@ class SwooleClient implements Client, ServesStaticFiles
             return;
         }
 
-        if ($length <= $this->chunkSize || config('octane.swoole.options.open_http2_protocol', false)) {
+        if ($length <= $this->chunkSize || config('octane.swoole.options.open_http2_protocol', false) || ! $canStream) {
             $swooleResponse->end($content);
 
             return;
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -226,8 +226,6 @@ class SwooleClient implements Client, ServesStaticFiles
             return;
         }
 
-        $canStream = config('octane.swoole.options.enable_coroutine', false);
-
         if ($octaneResponse->outputBuffer) {
             $swooleResponse->write($octaneResponse->outputBuffer);
         }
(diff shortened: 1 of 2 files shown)


taylorotwell:

L3o-pold:
For the record it was carefully human reviewed and rewritten by hand and not just AI generated. It also resolved a real use case we found in production and tested against a real bug we faced (again in production).

L3o-pold:
> If you're interested in contributing to Laravel, I'd encourage you to familiarize yourself with the existing codebase, engage with the community,

also thanks for the feedback but I was a contributor of both Laravel and Octane before AI was a thing.

dam1r89:
Hey, @L3o-pold did you test this patch in production env?
I see the same exception on my end and I would use this fix.

But I did not understand this: if I just enable coroutines flag, then this fix would not be necessary? Is there any downside of doing that?

L3o-pold:
@dam1r89 enabling coroutines is risky (and theoretically not supported, see related issue), an other dirty path (but not risky) is to enable `open_http2_protocol`.

dam1r89:
Thanks! So, setting `open_http2_protocol` to `true` is ha "hacky" way to force it to use `end` instead of `write` which requires coroutine.

PR #1155, opened 2026-08-04
Fix FrankenPHP output buffering

dkkoma:
This PR makes it possible to view the output of `dump()` in the browser when using FrankenPHP.

Viewing `dump()` output in the browser is already supported by `SwooleClient` and `RoadRunnerClient`. `dump()` remains a developer-friendly and convenient way to quickly inspect values during development. For that reason, it would be great to have the same functionality available with FrankenPHP.

Thank you for your consideration.


taylorotwell:

dkkoma:
Thanks for the response, and for the nudge toward looking outside the framework.

I went back and researched this again, and found Buggregator. It receives dump() output over the Symfony VarDumper protocol and shows it in its own UI, which solves the problem this PR was trying to address — and does it without Octane having to touch the response at all. That's a better fit than what I proposed here.

Thanks again!

PR #1163, opened 2026-08-23
Fix FrankenPHP generator stream responses

wakqasahmed:
Fixes #1142.

response()->stream() with a generator callback is silently broken under FrankenPHP. The Octane FrankenPHP client's respond() just delegates to Symfony's StreamedResponse::send(), which calls sendContent() — and sendContent() just invokes the callback, it never iterates a return value. It's written for callbacks that echo their own output. So when Laravel's ResponseFactory::stream() installs a generator directly as the callback (which it does whenever LARAVEL_OCTANE is set, see Illuminate\Routing\ResponseFactory::stream), FrankenPHP runs the generator, throws away everything it yielded, and sends a 200 with an empty body.

RoadRunnerClient already has to deal with this same mismatch and solves it by reflecting on the callback to see if it's a generator, and if so calling it directly and echoing each yielded chunk instead of handing it to Symfony. I did the same thing here for FrankenPhpClient: sendHeaders() up front (since we're bypassing send() entirely for this path), then iterate the generator with echo + flush per chunk.

There's an earlier attempt at this in #1141 with the same core idea. I kept the fix itself close to that, since it matches the existing RoadRunner convention, but simplified resolveStreamResponseCallback down to just the isGenerator() check — that's exactly what Laravel's own ResponseFactory uses to decide whether to hand FrankenPHP a raw generator in the first place, so checking for anything else (like an explicit `: string` return type) isn't actually reachable from that code path and just adds surface area.

For tests, I added a unit test that reproduces the bug directly against FrankenPhpClient::respond() with a generator-returning StreamedResponse, a test confirming plain echo-based streamed callbacks still fall through to Symfony's send() untouched, and an end-to-end test that boots a real worker with a route registered through response()->stream() (with LARAVEL_OCTANE set, same as production) and asserts the SSE-style output actually comes out the other end. I ran the new tests against the old respond() first to confirm they fail with an empty string (reproducing the reported bug exactly), then confirmed they pass with the fix, then ran the full suite to make sure nothing else broke.

### DIFF
--- a/src/FrankenPhp/FrankenPhpClient.php
+++ b/src/FrankenPhp/FrankenPhpClient.php
@@ -2,13 +2,16 @@
 
 namespace Laravel\Octane\FrankenPhp;
 
+use Generator;
 use Illuminate\Foundation\Application;
 use Illuminate\Http\Request;
 use Laravel\Octane\Contracts\Client;
 use Laravel\Octane\Octane;
 use Laravel\Octane\OctaneResponse;
 use Laravel\Octane\RequestContext;
+use ReflectionFunction;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\StreamedResponse;
 use Throwable;
 
 class FrankenPhpClient implements Client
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #1164, opened 2026-08-23
Fix Swoole coroutine error on responses over the chunk size

wakqasahmed:
Fixes #1150.

When a response body is bigger than the configured chunk size (1MB by default), SwooleClient sends it with a loop of `$swooleResponse->write()` calls followed by `end()`. That `write()` is Swoole's chunked-transfer write, and it needs an active coroutine to run. Octane's Swoole server sets `enable_coroutine => false` and doesn't wrap the request callback in a coroutine itself, so as soon as a response crosses the chunk threshold and hits that write loop, Swoole throws `API must be called in the coroutine`. Responses at or under the threshold go through `end($content)` instead, which doesn't have this requirement, so they've always worked fine — that's why this only shows up on larger responses.

The same write() calls exist for StreamedResponse output and for any leftover output buffer content, so those have the same latent issue even though nobody has reported it yet.

I didn't want to just swap the loop for a single `end($content)` call, since Swoole's http server buffers the whole body up to `package_max_length`/`buffer_output_size` (2MB by default) before that limit kicks in, so a large response would just fail a different way. The chunking is there for a reason.

Instead this wraps the write() calls in `Swoole\Coroutine\run()`, which gives them a real coroutine context without changing `enable_coroutine` globally or touching how the rest of the request is handled. This is the same pattern Octane already uses in SwooleCoroutineDispatcher for other coroutine-only Swoole APIs, so it's consistent with how the codebase handles this elsewhere.

I added a test that asserts the write() calls happen while `Swoole\Coroutine::getCid()` returns a real coroutine id (not -1) — it fails against the current code and passes with this fix. Ran the full test suite with the real swoole extension loaded (via the phpswoole/swoole Docker image) and everything passes, 185 tests.

### DIFF
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -12,6 +12,7 @@ use Laravel\Octane\Octane;
 use Laravel\Octane\OctaneResponse;
 use Laravel\Octane\RequestContext;
 use ReflectionClass;
+use Swoole\Coroutine;
 use Swoole\Http\Response as SwooleResponse;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
 use Symfony\Component\HttpFoundation\Response;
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #1165, opened 2026-08-26
Fix Swoole coroutine error on large responses

wakqasahmed:
Fixes #1150.

SwooleClient sends large response bodies (and StreamedResponse output) with a loop of `$swooleResponse->write()` calls, but those need a coroutine context to run and Octane's Swoole server doesn't provide one for the request callback — hence the "API must be called in the coroutine" crash once a response crosses the chunk size. Wrapped the write loops in `Swoole\Coroutine\run()`, which is the same pattern `SwooleCoroutineDispatcher` already uses elsewhere in this codebase for coroutine-only APIs. Added a test that checks the write calls happen inside a real coroutine (fails on current code, passes with the fix), and ran the full suite against php8.2 with the swoole extension — 188 passing.

### DIFF
--- a/src/Swoole/SwooleClient.php
+++ b/src/Swoole/SwooleClient.php
@@ -12,6 +12,7 @@ use Laravel\Octane\Octane;
 use Laravel\Octane\OctaneResponse;
 use Laravel\Octane\RequestContext;
 use ReflectionClass;
+use Swoole\Coroutine;
 use Swoole\Http\Response as SwooleResponse;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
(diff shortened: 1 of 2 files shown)


taylorotwell:

Cluster2a:
> Thanks for your pull request to Laravel!
> 
> Unfortunately, I'm going to delay merging this code for now. To preserve our ability to adequately maintain the framework, we need to be very careful regarding the amount of code we include.
> 
> If applicable, please consider releasing your code as a package so that the community can still take advantage of your contributions!

Totally fair. Can't wait to composer require octane/actually-finish-sending-the-response - the community deserves complete HTTP responses as an opt-in package.

